PR #10, opened 2022-11-27
Fill regions on 'F' key (https://github.com/Fish1/yapms-beta/issues/7)

### LINKED ISSUE #7: F Button Shortcut - Quick Fill
Holding the F button should allow the user to hover over regions to fill them in.

Flizzyclone:
To use:
Hold 'F'
Drag cursor across map
yapmsFill

Works on buttons as well.

Summary of code changes:
- Adds fillKeyPressed boolean variable to map page along with a get method.
- Adds handler methods for keyDown and keyUp events. Necessary to track whether F key is currently pressed.
- Adds methods to run onmouseover for states and map buttons.
- Changes fillRegion function in map page to no longer assume increment parameter is true, changes setupRegions and setupButtons methods in initialize.ts to accommodate. This is necessary to prevent behavior where mousing over a state twice while filling will "double fill" and change margin (see gif below).
yapmsDoubleFill

Closes #7.
Please let me know if you need more documentation.

### DIFF
--- a/client/src/routes/app/[country]/[map]/[year]/+page.svelte
+++ b/client/src/routes/app/[country]/[map]/[year]/+page.svelte
@@ -23,6 +23,7 @@
 	};
 
 	let mode: Mode = 'fill';
+	let fillKeyPressed: boolean = false;
 
 	let currentMap = 'usa' as keyof typeof imports;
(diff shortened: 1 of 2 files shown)


Fish1:
LGTM

PR #15, opened 2022-12-01
Disabled States 

Flizzyclone:
Adds disabled states, states are disabled by setting the "disabled" attribute on their object. This can be done beforehand in the svg (for maps like Senate classes) or in editor using the Disable function (gif below). Disabled states revert back to "Tossup" when re-enabled. Uses styles from yapms1.

yapmsDisabled

Summary of code changes:
- Adds 'disabled' as valid string for Mode type, adds optional disabled property to State interface.
- Creates new method _toggleRegion in logic. This function either disables or enables the state by setting/removing the disabled attribute, setting styles for both the state and any button associated, and setting fill to "Tossup" if disabling. Adds/removes state value from tossup margin to take state value "off the board".
- Modifies fillRegion function to make it do nothing if state to be filled is disabled.
- Modifies editRegion function to only update margins if region enabled.
- Modifies initializeMap to add disabled styles to states disabled in svg.

Tested both locally & using Docker


Fish1:
@Flizzyclone Looks nice, I'll try to merge it in soon.

Going forward, could you just make an issue in the issue tab before you start working on something? This way you can let people know you're working on something, and people aren't duplicating work without knowing.

Flizzyclone:
> @Flizzyclone Looks nice, I'll try to merge it in soon.
> 
> Going forward, could you just make an issue in the issue tab before you start working on something? This way you can let people know you're working on something, and people aren't duplicating work without knowing.

You got it, I'll make sure to make an issue in the future here.

PR #24, opened 2022-12-06
Fix to odd pie chart behavior

### LINKED ISSUE #22: Odd pie chart behavior
When a candidate's electoral votes move from or to 0, such as when the map is painted for the first time, the pie chart goes crazy.
yapmsCrazyPie

I think I have a fix for this, I'll test and create a PR if I find a working solution.

Flizzyclone:
Changes pie chart behavior to make transitions more smooth.
yapmsFixedPie

Why does this work?
Basically, passing undefined to the counts array doesn't add anything. This means the number of elements in the array is constantly changing. This means that the pie chart has to re-render in order to add a new element for the candidate's "slice".

If we have all the candidate slices render by keeping the counts array length == # of candidates, but just set the size of them to 0 if the candidate has 0 EV's, then the size of these slices can be readjusted instead of the chart re-rendering completely. Chart should in theory still re-render when a candidate is added or deleted since the counts array length will change.

Closes #22.
Tested using Docker & Local

### DIFF
--- a/client/src/lib/components/chartbar/ChartBar.svelte
+++ b/client/src/lib/components/chartbar/ChartBar.svelte
@@ -14,9 +14,9 @@
 		const labels: string[] = [];
 
 		$CandidatesStore.forEach((candidate) => {
-			const count = $CandidateCounts.get(candidate.id);
+			let count = $CandidateCounts.get(candidate.id);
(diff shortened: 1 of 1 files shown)


Fish1 (approved):
LGTM

PR #26, opened 2022-12-06
Quick fill fix

### LINKED ISSUE #23: Quick Fill Region Feature
Like Disabled States, quick-filling states got broken in the change to subscription stores. I'll fix this.

Flizzyclone:
Quick fill functionality originally implemented before store rework with PR #10 has been readded.
yapmsFill2

Summary of code changes:
- Creates store (InteractionStore) for keyboard state, uses obtained from [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
- Adds handler methods for keyDown and keyUp events. Necessary to track whether F key is currently pressed.
- Adds a method to run onmouseover for states and regions.
- Adds optional 'increment' parameter to fillRegion method to stop the "double fill" behavior described in the original PR.

Closes #23 
Tested on Docker & Local

### DIFF
--- a/client/src/lib/stores/Interaction.ts
+++ b/client/src/lib/stores/Interaction.ts
@@ -1,4 +1,4 @@
 import { writable } from 'svelte/store';
-/*The array here is an array of keys pressed by code (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code). 
-Limited to key presses relevant to functionality through checks where the store is modified.*/
-export const InteractionStore = writable<Array<String>>([]);
+/*The map here is a map of keys pressed by code (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code). 
+Key presses are set to true when pressed and then deleted from the map when released.*/
+export const InteractionStore = writable<Map<string, boolean>>(new Map<string,boolean>);
--- a/client/src/lib/stores/Interaction.ts
+++ b/client/src/lib/stores/Interaction.ts
@@ -1,0 +1,4 @@
+import { writable } from 'svelte/store';
+/*The array here is an array of keys pressed by code (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code). 
+Limited to key presses relevant to functionality through checks where the store is modified.*/
+export const InteractionStore = writable<Array<String>>([]);
--- a/client/src/routes/app/[country]/[map]/[year]/+page.svelte
+++ b/client/src/routes/app/[country]/[map]/[year]/+page.svelte
@@ -11,6 +11,8 @@
 	import ClearMapModal from '$lib/components/modals/clearmapmodal/ClearMapModal.svelte';
 	import applyPanZoom from './initialize/ApplyPanZoom';
 	import EditRegionModal from '$lib/components/modals/editregionmodal/EditRegionModal.svelte';
+	import { get } from 'svelte/store';
+	import { InteractionStore } from '$lib/stores/Interaction';
 
 	const imports = {
 		usa: () => import('$lib/assets/usa.svg?raw'),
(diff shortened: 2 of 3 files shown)


client/src/lib/stores/Interaction.ts
```diff
@@ -0,0 +1,4 @@
+import { writable } from 'svelte/store';
+/*The array here is an array of keys pressed by code (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code). 
+Limited to key presses relevant to functionality through checks where the store is modified.*/
+export const InteractionStore = writable<Array<String>>([]);
```
Fish1: I think this should be a 
```js 
Map<string, boolean>
```
this will allow for O(1) write and read speeds, instead of O(n) with the array.

Also, just as a rule of thumb, primitives should always be used over their object implementations. Array<String> should be string[] if you need an array of strings.

client/src/routes/app/[country]/[map]/[year]/+page.svelte
```diff
@@ -27,12 +29,33 @@
...
+		}
+	}
+
+	function handleKeyUp(e: KeyboardEvent) {
+		if (validKeyCodes.indexOf(e.code) !== -1) { //Check if key used for functionality
+			let interactions = get(InteractionStore);
+			interactions = interactions.filter(code => code !== e.code); //Remove code for key released
+			InteractionStore.set(interactions);
+		}
+	}
```
Comment (Fish1) [inline client/src/routes/app/[country]/[map]/[year]/+page.svelte]: These functions can both be simplified to one line of code.

```js
function handleKeyDown(e: KeyboardEvent) {
  $InteractionStore.set(e.code, true);
}

function handleKeyUp(e: KeyboardEvent) {
  $InteractionStore.delete(e.code);
}
```

While this might store more key strokes than we need, it will prevent us from having a conditional check for valid options. So I think the performance will be about the same.

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -4,8 +4,9 @@ import { RegionsStore } from '$lib/stores/Regions';
 import { TossupCandidateStore, SelectedCandidateStore } from '$lib/stores/Candidates';
 import { EditRegionModalStore } from '$lib/stores/Modals';
 import type Region from '$lib/types/Region';
+import { InteractionStore } from '$lib/stores/Interaction';
 
-function fillRegion(regionID: string) {
+function fillRegion(regionID: string, increment?: boolean) {
```
Comment (Fish1) [inline client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts:9]: I think in general, optional parameters should be avoided. It's fine for now, it just becomes a hassle to do more checks for possible undefined values as a project grow.

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -79,6 +80,15 @@ function loadRegions(node: HTMLDivElement) {
 			}
 		};
 
+		childHTML.onmousemove = () => {
+			const currentMode = get(ModeStore);
+			const currentInteractions = get(InteractionStore);
+
+			if(currentMode === 'fill' && currentInteractions.indexOf("KeyF") !== -1) {
```
Comment (Fish1) [inline client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Since this is a hash map, we can now just do...
```js
currentInteractions.get('KeyF') === true
```

Fish1 (changes requested):
The general idea for the setup looks good. Just a couple details like switching to a hash map for O(1) searches instead of linear might be better.

Not that were going to have a million keys pressed at any moment. lol

Fish1:
I said O(n) where I meant O(1), fixed now

client/src/routes/app/[country]/[map]/[year]/+page.svelte
Comment (Fish1) [reply client/src/routes/app/[country]/[map]/[year]/+page.svelte]: I was wrong about using the stores like I showed in my example. You should be using ```InteractionStore.set``` to update the store.

I'll need to go through and fix the situations where I did it incorrectly.

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Flizzyclone) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts:9]: I am in agreement, did not see another way to do it here. Breaking it out into a function is less desirable, and making it a required parameter requires changes to all the other calls, which we could do.

Flizzyclone:
@Fish1 please re-review. Changed to map & simplified functions based on above comments and info from Svelte Discord.

Fish1 (approved):
LGTM!

I think on the new map instantiation, we should keep the parenthesis. But I'll add a more strict ESLint config to catch more things like these.

PR #27, opened 2022-12-07
Add Candidate Functionality

### LINKED ISSUE #19: Ability to add new candidates
Original YAPms has a button next to the candidate list to add a new candidate.

Copy the modal, but consider adding a button the the navbar.

image

Fish1:
**Feature**

* added an "add candidate" button to the navbar, which opens a modal.
* created the "add candidate" modal.
* the modal has the ability to set a name, and pick colors.

image

image

image

**Code**
1) Created an "AddCandidateModal"
2) Imported the "AddCandidateModal" component into the main page.
3) Added a store to track the state of the "AddCandidateModal"

The add candidate modal will update the CandidatesStore on confirm.

It will add a new candidate to the CandidateStore via the user input.

closes #19

### DIFF
--- a/client/src/lib/components/modals/addcandidatemodal/AddCandidateModal.svelte
+++ b/client/src/lib/components/modals/addcandidatemodal/AddCandidateModal.svelte
@@ -1,0 +1,102 @@
+<script type="ts">
+	import { CandidatesStore } from '$lib/stores/Candidates';
+	import { AddCandidateModalStore } from '$lib/stores/Modals';
+
+	let newName = '';
+	let newColors: string[] = ['#000000'];
+
+	function initialize() {
+		newName = '';
+		newColors = ['#000000'];
+	}
+
+	function addColor() {
+		newColors = [...newColors, '#000000'];
+	}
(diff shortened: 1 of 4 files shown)


Flizzyclone:
All looks good to me, only thing I am curious about is why candidate ID's are assigned using crypto rather than just iterating by 1 from the current highest ID.

Flizzyclone (approved):

Fish1:
@Flizzyclone  I went with a UUID because you can generate a new ID from anywhere without worrying about dependencies.

With an incrementing ID you'd need to store a global variable somewhere, or calculate the next one by checking all the current IDs that exist in the candidate array.

Using the randomUUID function we can create a new UUID from anywhere without any checks.

Flizzyclone:
Ah yeah, makes sense. My idea was that you just check the last candidate in the array for the biggest ID but that has pitfalls if you delete out of order.

Good thinking, this is done really well. Perhaps we add in a list of colors to pick from at some point so it isn't just #000 every time.

PR #33, opened 2022-12-11
Fix buttons

### LINKED ISSUE #25: Buttons cannot be filled/edited.
Clicking the buttons on the side of map (with the exception of parts of the DC button not covered by text), doesn't do anything.

Fish1:
**Feature**

Added all the same interactions on the buttons that are on the regions.

DC has strange text, because it doesn't have a region associated with it.

image
image

**Code**

Moved the styles to the parent, because every region / button has the same pointer.
image

Simply assigning the onclick and onmousemove to the same region function works.
image

closes #25


Flizzyclone (approved):
LGTM

PR #39, opened 2022-12-11
Fixes to candidate editing

### LINKED ISSUE #38: Candidate colors do not update when changed
When you fill states for a candidate and then update their colors, the states filled on the map do not change.

Flizzyclone:
Regions not updating colors when candidate edited, selected candidate store not updating when candidate edited, and fixes the Tossup candidate being able to be opened in the edit modal.

### **Regions not updating colors when candidate edited (#38)**
When a candidate was edited, the versions of the candidate object that were present in the region store were not updated. Now, they are.
yapmsColorUpdate

### **Selected Candidate not updating when edited**
Previously, when you updated the selected candidate, SelectedCandidateStore would not update, and as such when you went to fill, you would fill with the old color.
yapmsOutdatedFill

This is fixed in this PR by updating SelectedCandidateStore when the currently selected candidate is edited.

### **Tossup candidate being able to be opened in the edit modal**
Despite not being editable, the Tossup candidate still had the enabled edit button next to it. 
image

This PR changes this by adding a an "editable" parameter when using the CandidateBox component. When set to true, this parameter disables the "Edit" button.
image

Additionally, there is a fix to a bug I made when readding quick fill where you could not increment margin when filling.

Closes #38.


client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -17,7 +17,7 @@ function fillRegion(regionID: string, increment?: boolean) {
 			count: region.value,
 			margin: 0
 		};
-		if (currentCandidate.candidate.id === selectedCandidate.id && increment === true) {
+		if (currentCandidate.candidate.id === selectedCandidate.id && increment === undefined) {
```
Comment (Fish1) [inline client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: This is a bit confusing, because now if you pass in true, it will not increment.

I would recommend to just make the parameter required. And then saying ```increment === true```

client/src/lib/components/modals/editcandidatemodal/EditCandidateModal.svelte
```diff
@@ -1,6 +1,8 @@
 <script lang="ts">
-	import { CandidatesStore } from '$lib/stores/Candidates';
+	import { CandidatesStore, SelectedCandidateStore, TossupCandidateStore } from '$lib/stores/Candidates';
```
Fish1: TossupCandidateStore doesn't need to be imported.

client/src/lib/components/modals/editcandidatemodal/EditCandidateModal.svelte
```diff
@@ -37,6 +39,18 @@
 			})
 		};
 		$EditCandidateModalStore.open = false;
+		//Update selected candidate object
+		if ($SelectedCandidateStore.id === id) {
+			$SelectedCandidateStore = $CandidatesStore[candidateIndex];
+		}
+		//Update candidate object in region store
+		let newRegions = get(RegionsStore);
```
Fish1: This should be a const

Fish1 (changes requested):
I think the increment parameter should be changed, to required. Because the logic now fails if you pass ```increment=true```.

Otherwise the rest looks good, just mentioned some linting things. I'll try to make the linter throw errors to point them out.

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Flizzyclone) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Thought about this earlier, held off. Will change.

Flizzyclone:
@Fish1 please re-review

Fish1 (approved):
LGTM

PR #40, opened 2022-12-12
Linting on Github Actions

Fish1:
- run the linter with --max-warnings=0
- this will result in an error on the PR to notify the user to fix issues


Fish1:
The lint pass or fail can be seen below.

image

PR #52, opened 2022-12-17
Reintroduction of Disabled States

### LINKED ISSUE #21: Disable Region Feature
After the rework, the disable feature needs to be recreated.

Global data will be in the stores now, instead of in the +page file. I think the state of the keyboard should also be handled in a store.

image

Flizzyclone:
Disabled States functionality originally implemented before store rework with #15 has been readded.
yapmsDisabledStatesReintro

Changes to feature:
When reenabled, the disabled state will now go back to it's previous fill. This is a deviation from YAPms1 and also creates the need for some potentially suboptimal solutions, such as setting the winner candidate to tossup in the function that handles styling regions. I made the determination this was ok, but it will be easy to reconfigure the feature to work as in YAPms1 if requested.

Summary of Code Changes:
- Creates a new property called "permaVal" within Region type. This value will track with "value" prop unless state is disabled, upon which value will be set to 0 and permaVal will not change. This is used to set state back to proper value when re-enabled. Can't just read this from map attributes in the case that user edit's region values.
- Creates new method disableRegion in initialize. This function either disables or enables the state by setting/removing the disabled attribute, setting styles for both the state and any button associated, and setting the region value to 0 if disabling, permaVal if enabling.
- Adds a disabled check to function that styles regions in order to enforce a consistent style.
- Adds a disabled check when editing region value in order to keep val at 0 if state disabled, edit region also now changes permaVal.

Closes #21.

### DIFF
--- a/client/src/lib/stores/Regions.ts
+++ b/client/src/lib/stores/Regions.ts
@@ -14,10 +14,16 @@ export const RegionsStore = writable<Region[]>([]);
 */
 RegionsStore.subscribe((regions) => {
 	regions.forEach((region) => {
-		let winner = region.candidates.reduce(
-			(prev, current) => (prev.count > current.count ? prev : current),
-			region.candidates[0]
-		);
+		const winner = region.disabled
+			? {
+					candidate: get(TossupCandidateStore),
+					count: 0,
+					margin: 0
+			  }
+			: region.candidates.reduce(
+					(prev, current) => (prev.count > current.count ? prev : current),
+					region.candidates[0]
+			  );
 		let marginIndex = winner.margin ?? 0;
 		if (marginIndex >= winner.candidate.margins.length) {
 			marginIndex = winner.candidate.margins.length - 1;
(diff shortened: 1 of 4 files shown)


Fish1 (changes requested):
(deleted)

client/src/lib/stores/Regions.ts
```diff
@@ -13,7 +14,7 @@ export const RegionsStore = writable<Region[]>([]);
 */
 RegionsStore.subscribe((regions) => {
 	regions.forEach((region) => {
-		const winner = region.candidates.reduce(
+		let winner = region.candidates.reduce(
 			(prev, current) => (prev.count > current.count ? prev : current),
 			region.candidates[0]
 		);
```
Fish1: We could use a ternary here to reduce some work, and keep const correctness.

image

Fish1 (changes requested):
One suggestion to keep const correctness. But otherwise, LGTM.

client/src/lib/stores/Regions.ts
Flizzyclone: Changed!

Fish1 (approved):
LGTM

PR #55, opened 2022-12-19
Defining disabled states now possible using map SVG

### LINKED ISSUE #54: Disabled States on map load
When reimplementing disabled states, I neglected to make it so that if a state was defined as disabled in the map SVG, it would have a 0 value and styles applied accordingly.

Flizzyclone:
When reintroducing disabled states the other day, I overlooked the need for certain region vars/style attributes to be different depending on whether or not the state was defined as disabled in the map SVG itself. This PR remedies this.

Summary of Code Changes:
- Moves application of disabled styles to the function that handles the rest of region styles rather than setting them directly in the function that disables the state.
- Changes the value parameter on state initialization to be set to 0 if state defined as disabled in SVG. No data is lost due to permaVal being set to value.

Disabled attribute on the SVG element can be set to anything, as long as it has the attribute, it is considered disabled.
Closes #54.

### DIFF
--- a/client/src/lib/stores/Regions.ts
+++ b/client/src/lib/stores/Regions.ts
@@ -48,8 +48,21 @@ RegionsStore.subscribe((regions) => {
 		}
 
 		region.nodes.region.style.fill = winner.candidate.margins[marginIndex]?.color;
-		if (region.nodes.button)
+		region.disabled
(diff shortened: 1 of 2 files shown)


Fish1 (approved):
LGTM

PR #56, opened 2022-12-19
Battle Chart

### LINKED ISSUE #49: Battle Chart
This is a popular chart type that needs to be added.

image

Fish1:
Feature: added a battle chart similar to original YAPms. It will breakout by the margins for each candidate.

Code:

1) Calculate the candidate points by margin, by adding another derived store.
2) Added down arrow icons.
3) Added Battle Chart.
   - put the counts into an arrary
   - calculate percentages to control the display
   - calculate the winning candidate to color in the arrow
4) Added Battle Chart Label
   - display the count of the margin
   - display the color of the margin
   - use the percentage to set the flex basis
5) Created a store to remember the selected chart type and switch between charts.

closes #49


Fish1:
Sorry, I removed a drop shadow from the arrow. And removed tailwind typography.

Flizzyclone (approved):
LGTM, the only thing I would recommend for the future is disabling the ability to click "Battle" in the chart type dropdown when more than two candidates are present.

PR #59, opened 2022-12-22
Locked States

### LINKED ISSUE #53: Locked States
Locked states act like disabled states in that they are resistant to fill. Instead of having value set to 0 though, they have their votes still count to the candidate they were last filled with.

Flizzyclone:
Adds in "Locked States" which are just resistant to filling. Using same styles as disabled for now, can distinguish by the fact that they still use candidate colors and have non-zero values. Probably a good idea to change styles in the future.
image
Locked ND & SD, Disabled MN.

Summary of Code Changes
- Adds new "Locked" property to regions, set to true if region has "locked" attribute in SVG.
- Adds 'lock' as a valid mode type.
- Disables fill if region has locked property
- Adds incredibly simple lock function, I don't think I can simply it further due to the way the region store is set up.

Closes #53.


Fish1 (approved):
LGTM

PR #61, opened 2022-12-23
Per-SVG candidates and SVG-defined winners

Flizzyclone:
Adds the ability to define candidates on a per-map basis in SVG by adding a "region-candidates" class <foreignObject> in <defs>. Candidates are defined thru JSON, separated by a line split:
image

If candidates not set in the map SVG, uses default store as defined in lib/stores/Candidates.ts

Also adds the ability to set winners for a state and their margin when the map is loaded using the "candidate-id" and "candidate-margin" attributes respectively. Code handles this with a very ugly looking conditional when setting the candidates property of a region. If attribute not defined, defaults to tossup candidate with margin 0.
image

This extends what we can do with maps by a lot. For example, U.S Senate seats colored by most recent winning margin (not included in commit).
image

No issue for this one. Started working to see if we could do it, proceeded to do it.

_"Your scientists were so preoccupied with whether they could, they didn’t stop to think if they should."_


Fish1:
Just a reminder, if you could create an issue for a PR before working on it.

Helps us coordinate what people are working on, and if you have questions or something you can ask in the issue without making a PR.

Fish1:
@Flizzyclone would you mind committing an example? Either your SVG in the picture, or making the presidential map use the feature?

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -73,6 +78,28 @@ function lockRegion(regionID: string) {
...
+		if (candidateString.trim() !== '') {
+			//If string has content
+			const newCandidate: Candidate = JSON.parse(candidateString);
+			newCandidates.push(newCandidate);
+		}
+	}
+	if (newCandidates.length !== 0) {
+		//If no candidates are defined in SVG, use generics defined in stores/Candidates.ts
+		CandidatesStore.set(newCandidates);
+	}
```
Comment (Fish1) [inline client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: This could be made into a more declarative & functional style if we use an array in the foreignObject. Also, we wouldn't need to worry about white space.

image

Flizzyclone:
@Fish1, changed usa.svg to use Per-SVG candidates and created a predetermined_regions_example.svg for predetermined region attribute implementation:
image
_Predetermined regions example_

Fish1 (changes requested):
Also, one graphical bug. When going to edit the candidate, all the colors appear black. Otherwise, everything seems to be functioning properly.

image

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Fish1) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: candidateString was a bad variable name for that mapping.

Comment (Fish1) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Also, what are your thoughts on using an attribute somewhere to store the JSON data in a string?

I think this would make it easier for people using an SVG editor to add the data. Because I'm not sure how valid it is to just store JSON data inside the foreignObjects element. Would something like inkscape or adobe even let you do that through their GUI?

image

Comment (Fish1) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Also, we could use Zod here to validate the structure of the data. And we can prevent someone from sending bad data/types to the system.

So instead of doing that mapping I suggested earlier, we could just verify the structure with Zod and then do a simple assignment.

Comment (Flizzyclone) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Inkscape has a very feature-rich XML editor that allows you to set an attribute on the SVG like above, yes. 
Going to look at using Zod.

Flizzyclone:
@Fish1 please re-review, updated with the following changes:
- Added zod, changed the way Candidate type is defined to be inferred from a zod schema. Used zod to perform validation on "candidates" attribute.
- Changed where candidates are defined to the "candidates" attribute on the SVG document itself. Can be edited with Inkscape.

After changes I could not reproduce the bug where all candidate colors are 0. Perhaps somewhere you passed an invalid hex code? Either way, please double check that it is gone.

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -73,6 +78,20 @@ function lockRegion(regionID: string) {
...
+	<svg
+	...
+   	candidates="[ 
+   		{&quot;id&quot;:&quot;0&quot;, &quot;name&quot;:&quot;Joe Custo Biden&quot;, &quot;margins&quot;: [{ &quot;color&quot;: &quot;#000055&quot; }, { &quot;color&quot;: &quot;#000099&quot; }, { &quot;color&quot;: &quot;#0000ff&quot; }]},
+   		{&quot;id&quot;:&quot;1&quot;, &quot;name&quot;:&quot;Donald Trump&quot;, &quot;margins&quot;: [{ &quot;color&quot;: &quot;#550000&quot; }, { &quot;color&quot;: &quot;#990000&quot; }, { &quot;color&quot;: &quot;#ff0000&quot; }]} 
+   	]"
+	*/
+	const candidatesStringified = node.querySelector('svg')?.getAttribute('candidates'); //This doesn't return SVG other than the map SVG
+	const candidates = candidatesStringified != null ? JSON.parse(candidatesStringified) : null; //If candidate property not set, set candidates to null so the next check knows to use default candidates.
+	if (candidates !== null) CandidatesStore.set(CandidateStoreSchema.parse(candidates)); //If no candidates are defined in SVG, use generics defined in stores/Candidates.ts
```
Comment (Fish1) [inline client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: 1) this crashes if there is invalid JSON because JSON.parse with throw an error
2) this will also crash if your CandidateStoreSchema isn't matched

I was able to just throw everything in a try catch, and it will let the function finish executing if the data is incorrect.

image

Fish1 (changes requested):

client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Fish1) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: The black color error happened when I was using 3 letter hex format. Like #FFF. Maybe we can look into fixing that in a later PR because it doesn't have to do with this feature.

Comment (Flizzyclone) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Wouldn't we want that "crashing" behavior to tell the mapmaker that their data format is incorrect? I suppose console.error gives the same debug info, but I left it this way since the only time invalid input would be given is when someone is making a map.

Comment (Fish1) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: I think displaying an error in the console, or a popup saying something went wrong would provide the user with a better experience.

Without catching the errors, it made the map some weird red color.

Comment (Flizzyclone) [reply client/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Alright, lets go with the try catch then.

Flizzyclone:
Added try catch for parsing data

Fish1 (approved):
LGTM

PR #65, opened 2022-12-24
Preset colors

### LINKED ISSUE #50: Build in color schemes
Old YAPms had built in color schemes available for users to select from.

image
image
image

Fish1:
Feature: When preset colors is clicked, it will create a popup. This allows users to select from a list of pre-made colors.

Let me know if you think this would look better as a regular popup, instead of sliding up from the bottom. It would be a simple change.

image

Code:
1) Created a PresetColorsModal, with it's global storage.
2) Made the AddCandidateModalStorage also store the colors, this allows the colors to be changed from outside the modal.

closes #50


Flizzyclone:
I think where it might be valuable to have it be a separate window in the center is if we wanted to add more presets and have it scroll. Besides that, this all looks good.

Flizzyclone (approved):

PR #70, opened 2022-12-24
Fixes to map loading

### LINKED ISSUE #66: Maps will not load if no buttons or texts present
If no .region-buttons group or .region-texts group is defined in a map, the map won't load. Easy fix:
image

### LINKED ISSUE #67: panZoom doesn't apply to maps
panZoom doesn't apply to maps with an id other than "testing-map".

since node is a div containing only the map svg, we can do the following:
image

Flizzyclone:
Fixes two issues related to loading maps, one where the map won't load if no text or buttons are defined, and one where panZoom won't properly apply to maps without id #testing-map.

Summary of Code Changes:
- Changes applyPanZoom to not look for a map specifically called #testing-map and instead look for an svg element in general.
- Changes panZoom max zoom to 100 from 2, will be useful for more detailed maps like the US house.
- Adds two conditionals to check if there are any buttons or texts before running code that requires them.

Closes #66 and #67


Fish1 (approved):
LGTM

PR #82, opened 2022-12-26
Remove Candidate

### LINKED ISSUE #63: Remove Candidate
The user needs the ability to remove a candidate.

Fish1:
Feature: Ability to remove candidates from the list.

image
image

Code:
1) Added function to remove candidate to editcandidate modal.
2) Update the region store by removing all candidates equal to the deleted one.

closes #63

### DIFF
--- a/apps/yapms/src/lib/components/modals/editcandidatemodal/EditCandidateModal.svelte
+++ b/apps/yapms/src/lib/components/modals/editcandidatemodal/EditCandidateModal.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { CandidatesStore, SelectedCandidateStore } from '$lib/stores/Candidates';
+	import { CandidatesStore, SelectedCandidateStore, TossupCandidateStore } from '$lib/stores/Candidates';
 	import { RegionsStore } from '$lib/stores/Regions';
 	import { EditCandidateModalStore } from '$lib/stores/Modals';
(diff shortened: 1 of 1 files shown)


apps/yapms/src/lib/components/modals/editcandidatemodal/EditCandidateModal.svelte
```diff
@@ -23,6 +23,25 @@
 		newColors = newColors.slice(0, newColors.length - 1);
 	}
 
+	function removeCandidate() {
+		$CandidatesStore = $CandidatesStore.filter((candidate) => candidate.id !== id);
+		$SelectedCandidateStore = $TossupCandidateStore;
```
Flizzyclone:
```suggestion
		if ($SelectedCandidateStore.id === id) $SelectedCandidateStore = $TossupCandidateStore;
```
This is the only potential improvement I can pinpoint, making it so selected candidate only changes to Tossup when the selected candidate is the one being deleted.

Flizzyclone (approved):
LGTM with the exception of the little tweak above. Good to merge with or without imo.

apps/yapms/src/lib/components/modals/editcandidatemodal/EditCandidateModal.svelte
Fish1: Good idea, fixed!

PR #86, opened 2022-12-29
Authentication

### LINKED ISSUE #57: Backend Setup
I think pocket-base would be a good solution for hosting a back-end.

It gives us file-storage and authentication methods.

https://pocketbase.io/

Fish1:
Feature: added email & social authentications

1) register you account with an email
2) verify your email
3) login using email and password

OR

1) use social login

register

Code:
Created a new modal for auth. This modal has 3 possible components, login, register and social form components. Each one is responsible for handling their own auth.

Had to connect to a google project for google oauth2 and an SMTP server for email. This can be done in pocketbase.

Task Remaining:

- [ ] make failed auth popups
- [ ] popup to notify that verification email was sent

closes #57


Fish1:
Sorry about all those commits, I am not sure of a better way to fix the github CI scripts.

Also, let me know if you need SMTP or Google secret keys to setup email/google login.

Flizzyclone:
I am having issues with getting env variables on my end it looks like
image

This was because .env.example needed to be copied and renamed to .env on my end

Flizzyclone:
I also can't send email, missing the tool to do so it looks like:
image

Fish1:
> I am having issues with getting env variables on my end it looks like image
> 
> This was because .env.example needed to be copied and renamed to .env on my end

Just copy the env.example file, then insert the URL for your pocketbase instance.

Fish1:
> I also can't send email, missing the tool to do so it looks like: image

Yes, you'll need to connect pocketbase to an SMTP server to send email. I setup an email address I can give you the authentication too so you can test it. Or you can find an alternative if you have one.

apps/yapms/src/lib/components/modals/authmodal/LogoutForm.svelte
```diff
@@ -0,0 +1,13 @@
+<script>
+  import { PocketBaseStore } from "$lib/stores/PocketBase";
+</script>
+
+<div class="flex flex-col gap-2">
+  <h3>Welcome {$PocketBaseStore.authStore.model?.email}</h3>
```
Flizzyclone, line 6: I think this should be the user's username rather than email.

Flizzyclone (approved):
I have faith that the email works, don't need to test it. All seems to be good here. Only added one tiny suggestion.

apps/yapms/src/lib/components/modals/authmodal/LogoutForm.svelte
Fish1: Agreed, I'll change that.

PR #89, opened 2023-01-05
Button fixes

### LINKED ISSUE #87: 'Home' Button doesn't work
When in app, the Home button doesn't redirect.

### LINKED ISSUE #88: Clicking anywhere on home social buttons does not work
Due to an improperly placed <a>, only clicking on text for the social buttons on the home page redirects you properly. This is fixed on other buttons.

Flizzyclone:
Fixes both the home button not redirecting and buttons on the home page only working when you click on text. Also fixes home page button wrapping issues identified in the Home Page PR.

Summary of code changes
- Switches the structure of home social buttons to have the <a> completely wrapping the element instead of the <a> inside the button.
- Makes home social buttons nowrap, fixing wrapping issues.
- Wraps the "Home" button on the app page in an <a> redirecting to "/".

Closes #87 and #88

### DIFF
--- a/apps/yapms/src/lib/components/homepage/HomePageSidebar.svelte
+++ b/apps/yapms/src/lib/components/homepage/HomePageSidebar.svelte
@@ -10,30 +10,22 @@
 	</div>
 	<div class="mb-4">
 		<div class="divider">Social Links</div>
-		<div class="flex flex-wrap justify-evenly gap-2">
-			<a class="w-[45%]" href="https://www.reddit.com/r/YAPms/">
-				<button type="button" class="btn btn-sm w-full flex-nowrap">
-					<Fa icon={faReddit} class="pr-1"/>
-					Reddit
-				</button>
+		<div class="grid grid-cols-2 gap-2">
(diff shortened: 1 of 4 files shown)


Fish1 (changes requested):
Looking at the buttons more, I think we can clean it up a bit by removing the button component. And just give the <a> tag the button styling. (for the navbar and social links)

Also, when possible, the parent element should control the positioning & size of the child elements. This makes it much easier to turn the children into their own components and be used in other places.

For example, in the future we might wanna make the social link buttons into their own components. If they have the 45% width, then wherever we put them, they are forced at 45% width. But with these changes, we can put them in other places cand control their widths and layout via a parent component.

EDIT: (the discord flex-wrap should be flex-nowrap)

image

Flizzyclone:
Good catch, this is helped by the nowrap. Before without percentages they would wrap undesirably, hence the 45%. Good idea with the <a> replacing <button> as well. I will change

Flizzyclone:
Replaced all buttons with <a>'s in them to just be <a>'s. Please re-review

Fish1 (approved):
LGTM

PR #90, opened 2023-01-06
"class" attribute on regions changed to "region"

### LINKED ISSUE #71: Map SVG format update
The SVG map format should be finalized.

Should be able to identify regions via custom attributes, not the class attribute.

Flizzyclone:
Change from using the class attribute to identify regions to the unique region attribute as discussed in #71.

Summary of Code Changes:
- Changes the way the region class "id" prop is assigned to use the region attribute.
- Changes the way region nodes are assigned from using the class attribute to the region attribute.
- Updates all maps in file to use this new attribute.

Closes #71.


Fish1 (approved):
I like it, make sense!

PR #93, opened 2023-01-08
Remove sidebar whitespace

### LINKED ISSUE #92: Fix whitespace gap between map and sidebar
There is a white space gap, that causes the map to look like it's being cut off. It should be removed.

image

Fish1:
Preview:

image

Code:

Use the "display:none" and "display:flex" properties, because the divider-horizontal would take priority over the width and margin if it had media queries.

closes #92


Flizzyclone (approved):
LGTM

PR #103, opened 2023-01-16
Home Page updates from Pocketbase

### LINKED ISSUE #99: Load Updates from API
Updates on the sidebar of the home page should be stored in a Pocketbase collection and then the latest updates should be requested from pocketbase and rendered in the side bar.

Flizzyclone:
This PR changes the updates page on the sidebar of the home page to use updates provided by an "updates" collection in Pocketbase. This collection must be setup like such: 
image
image
(avoid a 403)

The updates sidebar fetches the 10 most recent updates and renders them like such:
image

yapmsDynamicUpdates

Code changes: 
- Adds new component SideBarUpdate consisting of the appropriate DOM elements for an update.
- Adds a function onMount for the HomePageSidebar component that fetches the 10 latest updates from Pocketbase.

### DIFF
--- a/apps/yapms/src/lib/components/homepage/HomePageSidebar.svelte
+++ b/apps/yapms/src/lib/components/homepage/HomePageSidebar.svelte
@@ -10,22 +10,24 @@
 	let updates:Record[] = [];
 
 	onMount(async () => {
-		//Request the last 10 update records from pocketbase.
-		$PocketBaseStore.collection('updates').getList(1, 10, {sort: '-created'})
-		.then(data => {
-			updates = data.items
-		}).catch(error => {
+		try {
+			//Request the last 10 update records from pocketbase.
+			const records = await $PocketBaseStore.collection('updates').getList(1, 10, {sort: '-created'});
+			updates = records.items;
+		} catch(error) {
 			console.log(`Failed to fetch updates:\n${error}`);
-		});
+		}
 	});
 </script>
 
 <div class="lg:w-1/5 pl-5 hidden md:flex flex-col h-full justify-between">
 	<div class="overflow-y-auto">
 		<div class="divider">Updates</div>
-		{#each updates as update}
-			<SideBarUpdate title={update.title} description={update.description}></SideBarUpdate>
-		{/each}
+		<div class="flex flex-col gap-y-3">
+			{#each updates as update}
+				<SideBarUpdate title={update.title} description={update.description}></SideBarUpdate>
+			{/each}
+		</div>
 	</div>
 	<div class="mb-4">
 		<div class="divider">Social Links</div>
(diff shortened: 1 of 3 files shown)


apps/yapms/src/lib/components/homepage/SideBarUpdate.svelte
```diff
@@ -0,0 +1,13 @@
...
+</script>
+
+<div class="card card-bordered bg-base-300 shadow-md p-3 mb-3">
+    <div class="text-xl text-center md:text-left font-semibold pb-2">
+        {title}
+    </div>
+    <div class="text-md text-center md:text-left">
+        {description}
+    </div>
+</div>
```
Fish1: *Requests*

1) Could you remove the mb-3 from the card, and wrap the #each loop inside a flex box with gap-3? The parent should be controlling the layout, not the children.


*Suggestions*

1) I think here you could use the daisyUI card-body and card-title classes to control the padding and font, if you want, other wise looks good.

2) I think the md:text-left and text-center can be removed, because the sidebar just disappears before it can shrink below medium.

apps/yapms/src/lib/components/homepage/HomePageSidebar.svelte
```diff
@@ -1,12 +1,31 @@
...
+
+	onMount(async () => {
+		//Request the last 10 update records from pocketbase.
+		$PocketBaseStore.collection('updates').getList(1, 10, {sort: '-created'})
+		.then(data => {
+			updates = data.items
+		}).catch(error => {
+			console.log(`Failed to fetch updates:\n${error}`);
+		});
+	});
```
Fish1: Request:

1) You don't need to use the async/await syntax if you're using callback based promises.

But I do think this should be converted to fully use the async/await syntax with a try/catch.

Avoid using the callback style unless it's really needed.

apps/yapms/src/lib/components/homepage/HomePageSidebar.svelte
```diff
@@ -1,12 +1,31 @@
...
 </script>
 
 <div class="lg:w-1/5 pl-5 hidden md:flex flex-col h-full justify-between">
-	<div>
+	<div class="overflow-y-auto">
 		<div class="divider">Updates</div>
-		<p>This will eventually be updated by a server</p>
+		{#each updates as update}
+			<SideBarUpdate title={update.title} description={update.description}></SideBarUpdate>
+		{/each}
```
Fish1: This should be wrapped in a flex box with a gap.

Fish1 (changes requested):
The pocketbase setup looks good to me. If you could just export the collections and put the JSON in a config or setup folder.

You can then use that JSON and import it into a new pocketbase setup.

image

apps/yapms/src/lib/components/homepage/SideBarUpdate.svelte
Flizzyclone: Request #1 and suggestion #2 implemented

apps/yapms/src/lib/components/homepage/HomePageSidebar.svelte
Flizzyclone: Ah yeah, this is a relic from when I was experimenting with a different implementation. Reworked to a try catch in latest commit.

Fish1 (approved):
LGTM

PR #109, opened 2023-02-13
Load maps based on URL

### LINKED ISSUE #105: Importing maps based on URL
We need a way to load different maps based on what page you are going to.

This will involve changing maps to a standardized name format following [country]-[map]-[year] like we use in the URL. Also will need a default fallback (2024 pres more than likely)

Flizzyclone:
This PR changes the loading of maps to be based off of URL slugs. To do this, all file names for maps now should follow the following format:
[country]-[map]-[year].svg     *e.x: usa-presidential-2022.svg*

You can go to localhost:8081/app/predetermined/regions/example.svg to load the predetermined regions example and verify that this works.

Code Changes:
- Moves all maps from /src/lib/assets to /src/lib/assets/maps, done mainly to keep glob importing from importing any other SVGs we may put into /assets
- Adds a load function in +page.server.ts that sends params to +page.svelte as "data".
- Adds functionality into +page.svelte to check for a map corresponding to URL slug and loads that map if it is valid. If that map does not exist, it will automatically load usa-presidential-2022.
- Fixes an issue where any URL other than /usa/presidential/2022 would be redirected back to /usa/presidential/2022 if not loading JSON.

Closes #105


apps/yapms/src/routes/app/[country]/[map]/[year]/+page.server.ts
```diff
@@ -1 +1,8 @@
+import type { PageServerLoad, PageServerLoadEvent } from './$types';
+
 export const prerender = true;
+
+/** @type {import('./$types').PageServerLoad} */
+export const load: PageServerLoad = ({ params }: PageServerLoadEvent) => {
+	return params;
+};
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/+page.server.ts]: This shouldn't be used, I think the application should run on the client side as much as possible.

apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte
```diff
@@ -26,13 +26,25 @@
...
 	import LoadingErrorModal from '$lib/components/modals/loadingerrormodal/LoadingErrorModal.svelte';
 
-	const imports = {
-		usa: () => import('$lib/assets/usa.svg?raw'),
-		nz: () => import('$lib/assets/nz.svg?raw')
-	};
+	/** @type {import('./$types').PageData} */
+	export let data;
+	//Data is URL slug information in the following format:
+	//{country:'usa',map:'presidential',year:'2022'}
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte]: We don't need to load the URL data from the server.

You can get the URL from the client like this.

import $page from "$app/stores";
const url = $page.url.pathname;

apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte
```diff
@@ -26,13 +26,25 @@
...
+
+	//Glob import is necessary here. If we try to use a template literal to import only one map, we are met with an error. 
+	//The query section is there to make sure the SVG contents are imported raw, seeing as you cannot glob import with .svg?raw
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte]: I think comments can be a little more succinct, they should explain what is happening. You don't need to explain why you did what you did, unless it's super necessary.

apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte
```diff
@@ -103,9 +115,10 @@
 
 			<div class="overflow-hidden w-full h-full">
 				<CandidateBoxContainer />
-				{#await imports[currentMap]() then module}
-					<div use:setupMap class="overflow-hidden h-full">
-						{@html module.default}
+				{#await imports[currentMap]()}
+					<h1>Loading Map...</h1>
+				{:then importedMap}<div use:setupMap class="overflow-hidden h-full">
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte]: div should be on a new line

apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte
```diff
@@ -26,13 +26,25 @@
...
+	const imports = import.meta.glob<typeof import("*?raw")>('$lib/assets/maps/*.svg', {
+		query: { raw:'' },
+	});
+
+	console.log(imports)
+	
+	//Make sure that even if map requested doesn't load, something loads.
+	let currentMap = '/src/lib/assets/maps/usa-presidential-2022.svg'; 
+	//If the map defined by slugs is found, use that map
+	if (imports[`/src/lib/assets/maps/${data.country}-${data.map}-${data.year}.svg`] !== undefined) { currentMap = `/src/lib/assets/maps/${data.country}-${data.map}-${data.year}.svg`; }
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte]: This is a bit long, can it be put on multiple lines?

apps/yapms/src/routes/app/+layout.svelte
```diff
@@ -14,7 +14,7 @@
 		const map = encodeURIComponent($page.url.searchParams.get("m") ?? "");
 		// if the user is not loading a map, redirect them to the default map
 		if (map === "") {
-			await goto(`/app/usa/presidential/2022`);
+			await goto($page.url);
 			return;
 		}
```
Fish1: I think we need this so the /app url gets redirected to something if the m= is missing. But actual pages don't get redirected.
```ts
if (map === "") {

	if ($page.url.pathname === "/app") {
		await goto("/app/usa/presidential/2022");
	}

	return;
}
```

apps/yapms/src/routes/app/+layout.svelte
```diff
@@ -29,7 +29,7 @@
 			LoadingErrorModalStore.set({
 				open: true,
 			});
-			await goto(`/app/usa/presidential/2022`);
+			await goto($page.url);
```
Fish1: This can go back to redirecting to ```await goto("/app/usa/presidential/2022");``` because it should be the default if something goes wrong.

apps/yapms/src/routes/app/[country]/[map]/[year]/+page.server.ts
Comment (Fish1) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/+page.server.ts]: Just leave the ```export const prerender = true```

apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte
```diff
@@ -26,13 +26,25 @@
...
+	//Data is URL slug information in the following format:
+	//{country:'usa',map:'presidential',year:'2022'}
+
+	//Glob import is necessary here. If we try to use a template literal to import only one map, we are met with an error. 
+	//The query section is there to make sure the SVG contents are imported raw, seeing as you cannot glob import with .svg?raw
+	const imports = import.meta.glob<typeof import("*?raw")>('$lib/assets/maps/*.svg', {
+		query: { raw:'' },
+	});
+
+	console.log(imports)
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/+page.svelte]: Shouldn't have console logs.

Fish1 (changes requested):
Just need to convert it to run fully client side.

Flizzyclone:
Changed to use page store and therefore run clientside... looks like we have some merge conflicts to be fixed when we merge in, so not much I can do for now.

Fish1 (approved):
LGTM. I just removed the package-lock.json

PR #118, opened 2023-02-28
Home page theme

### LINKED ISSUE #115: Home Page Theme Stuck
The theme on the home page doesn't change when it's changed in the app.

I think it should be as simple as adding this function on mount.

image

Flizzyclone:
This PR solves the issue of themes not being applied to the home page & adds a "Theme" button and modal to the home page for switching themes.
image
image
image

Summary of Code Changes
- Adds themeChange(false) to the home page to run onMount
- Adds HomeThemeModal, similar code to style modal with chart position removed.
- Adds ThemeGrid component to share the button grid for selecting themes between components. This is ideal due to the way the grid is constructed from an array- we don't want to have to update that array in multiple files.
- Adds the Swatch icon
- Adds a desktop & mobile variant for the theme button to the home page.


Fish1 (changes requested):
I think the HomeThemeModal and HomeThemeModalStore should remove the Home part.

And maybe the ThemeGrid could be moved to the ModalUtilities folder for now.

Otherwise looks good!

Flizzyclone:
Made changes as requested, please re-review.

Fish1 (approved):
LGTM

PR #123, opened 2023-03-02
Adding Japan

EllieSD:


Flizzyclone:
A couple things I notice here- text for insets does not react to the background. This is a problem me and Fish have to solve programmatically.
image

Secondly, the map is using default candidates still. You can actually implement candidates within the map. This is more technical, so if you want me to do it I will. You can find an example for how to do this in predetermined-regions-example.svg:
```
candidates="[ 
      {&quot;id&quot;:&quot;0&quot;, &quot;name&quot;:&quot;Joe Custom Candidate Biden&quot;, &quot;margins&quot;: [{ &quot;color&quot;: &quot;#000055&quot; }, { &quot;color&quot;: &quot;#000099&quot; }, { &quot;color&quot;: &quot;#0000ff&quot; }]},
      {&quot;id&quot;:&quot;1&quot;, &quot;name&quot;:&quot;Donald Unique&quot;, &quot;margins&quot;: [{ &quot;color&quot;: &quot;#550000&quot; }, { &quot;color&quot;: &quot;#990000&quot; }, { &quot;color&quot;: &quot;#ff0000&quot; }]} 
]">
```

It's a bit more comprehendible in Inkscape:
image

This candidates property is added to the svg element itself and follows the format:
```
[
{"id":"0","name":"Example Name","margins":[{"color":"#HEX CODE"},{"color":"#HEX CODE"}]},
{"id":"1","name":"Example Name","margins":[{"color":"#HEX CODE"},{"color":"#HEX CODE"}]}
]
```

Fish1:
I think we should remove the default tooltips, because they look pretty ugly. Maybe do something that looks better.

Also I think the square boxes can be deleted. I don't know how much value they add to the map, and they stand out a lot.
image

Also can you make the map name lowercase?
image

I tried pushing to the branch myself, but I don't think I have permission.

Fish1 (changes requested):
I think we should remove the default tooltips, because they look pretty ugly. Maybe do something that looks better.

Also I think the square boxes can be deleted. I don't know how much value they add to the map, and they stand out a lot.
image
image


Also can you make the map name lowercase?
image

I tried pushing to the branch myself, but I don't think I have permission.

Jackess4334:
The boxes should be removed and the cities/zoomed regions should be easier to read

Fish1:
Closing this out unless someone would like to update the Japan map to work.

PR #124, opened 2023-03-03
Input validation for edit region modal

### LINKED ISSUE #101: Editing region value to a negative number bricks the app.
As described in the title, editing a region to have a negative value bricks the app. Easy to fix w/ a simple conditional.

Flizzyclone:
This PR disables the confirm button in the edit region modal when an invalid input is given for the number of electors.

image
image

When requirements are not met, user can see what is wrong by hovering over the input. I can't capture this in a screenshot.

Summary of Code Changes:
- Adds a new property, confirmDisabled to EditRegionModalStore. This is set to false when the modal is closed, and is only set to true when an invalid input is entered.
- Adds required to the input element in order to not allow null values
- Adds an on:input function to control the value of confirmDisabled. The checkValidity() function will return false for anything that is not a non-negative integer.

Closes #101

### DIFF
--- a/apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
+++ b/apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
@@ -4,7 +4,6 @@
 	import ModalTitle from '../../modalutilities/ModalTitle.svelte';
 
 	$: open = $EditRegionModalStore.open;
-	$: confirmDisabled = $EditRegionModalStore.confirmDisabled;
 	$: longName = open ? $EditRegionModalStore.region?.longName : undefined;
 	$: value = open ? $EditRegionModalStore.region?.permaVal : undefined;
 	$: newValue = value ?? 0;
@@ -37,14 +36,11 @@
 			class="input input-bordered w-full"
 			min="0"
 			required
-			on:input={(event) => {
-				$EditRegionModalStore.confirmDisabled = !event.currentTarget.checkValidity();
-			}}
 			bind:value={newValue}
 		/>
 		<div class="modal-action">
 			<button class="btn btn-primary" on:click={close}> No </button>
-			<button class="btn btn-success" on:click={confirm} disabled={confirmDisabled}>
+			<button class="btn btn-success" on:click={confirm} disabled={newValue === null || newValue < 0 || newValue % 1 !== 0}>
 				Update
 			</button>
 		</div>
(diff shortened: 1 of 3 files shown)


apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -47,7 +47,8 @@ function editRegion(regionID: string) {
 	const region = get(RegionsStore).find((region) => region.id === regionID);
 	EditRegionModalStore.set({
 		region: region ?? null,
-		open: region !== undefined
+		open: region !== undefined,
+		confirmDisabled: false
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts:51]: I don't think this needs to be controlled globally from a store, it could be controled inside the modal itself.

apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
```diff
@@ -34,12 +35,18 @@
...
+			required
+			on:input={(event) => {
+				$EditRegionModalStore.confirmDisabled = !event.currentTarget.checkValidity();
+			}}
 			bind:value={newValue}
 		/>
 		<div class="modal-action">
 			<button class="btn btn-primary" on:click={close}> No </button>
-			<button class="btn btn-success" on:click={confirm}> Update </button>
+			<button class="btn btn-success" on:click={confirm} disabled={confirmDisabled}>
```
Fish1: I think what you could do here is simply check if the newValue is out of bounds, set disabled to true.

apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
```diff
@@ -4,6 +4,7 @@
 	import ModalTitle from '../../modalutilities/ModalTitle.svelte';
 
 	$: open = $EditRegionModalStore.open;
+	$: confirmDisabled = $EditRegionModalStore.confirmDisabled;
 	$: longName = open ? $EditRegionModalStore.region?.longName : undefined;
 	$: value = open ? $EditRegionModalStore.region?.permaVal : undefined;
 	$: newValue = value ?? 0;
```
Fish1: Maybe have a dynamic "newValueOutOfBounds" variable here that does the logic for checking outOfBounds.

Fish1:
```js
$: newValueOutOfBounds = newValue < 0;
```

Fish1 (changes requested):
This shouldn't be a global store, all the information needed for the logic is contained within the modal.

Flizzyclone:
Reworked the check to disable the "Confirm" button to be based on newValue. Could have this be a dynamic variable if you want, just lmk. @Fish1

Fish1 (approved):
LGTM

PR #126, opened 2023-03-03
Add data-sveltekit-reload to Home button

### LINKED ISSUE #125: Home Button doesn't properly redirect
When you hit the "Home" button, the animation to go back to home plays and the URL changes, but the page itself doesn't actually change.

Flizzyclone:
This PR simply adds data-sveltekit-reload to the home button so that the page reloads when the button is pressed.

As explained to me on the Svelte server, the issue where the page was not reloading was because the home page does not rely on any load methods.

Closes #125


Fish1 (changes requested):
This causes a full page refresh, it should rely on the client side routing.

PR #133, opened 2023-03-25
Server Side Screen Shots

### LINKED ISSUE #132: Screenshots
Screenshots should be done with browserless on the server.

Fish1:
Feature: When the user uploads their map, it will send a request to browserless to take a screenshot of it.

aaaa

Code:
1. added a screenshot to the map record
2. added browserless as docker container
3. implemented pocketbase logic to connect to browserless and initiate the screenshot
4. re-organized the env files so they can all be maintained in the same spot
5. added a dev container, that will install everything needed to start yapms

Browserless will auto start inside the devcontainer, otherwise it needs to be started manually.

Thinks to be broken out into other issues.
1. remove the top and sidebar from the screenshot
2. refactor the golang, break out code into functions

closes #132


Fish1:
Also, deployments wont work with browserless anymore

apps/pocketbase/main.go
```diff
@@ -44,6 +56,18 @@ func main() {
 	})
 
 	app.OnBeforeServe().Add(func(e *core.ServeEvent) error {
+
+		e.Router.AddRoute(echo.Route{
```
Flizzyclone, line 60: Did you forget to remove this route?

apps/pocketbase/main.go
```diff
@@ -134,22 +166,55 @@ func main() {
...
+				ctx, cancel := chromedp.NewContext(allocatorContext)
+				defer cancel()
+
+				// create a new browserless task
+				var screenshot []byte
+				if err := chromedp.Run(ctx,
+					chromedp.EmulateViewport(1600, 900),
+					chromedp.Navigate(browserlessFrontendURI + "/app?m=" + record.Id),
+					chromedp.WaitReady("#testing-map"),
+					chromedp.FullScreenshot(&screenshot, 100),
```
Flizzyclone, line 174: I know you said that this would be broken apart into another issue, but I found an example of a single element screenshot in [chromedp examples](https://github.com/chromedp/examples/blob/master/screenshot/main.go).

`elementScreenshot(`https://pkg.go.dev/`, `img.Homepage-logo`, &buf)`

```
func elementScreenshot(urlstr, sel string, res *[]byte) chromedp.Tasks {
	return chromedp.Tasks{
		chromedp.Navigate(urlstr),
		chromedp.Screenshot(sel, res, chromedp.NodeVisible),
	}
}
```

apps/pocketbase/migrations/1679449550_updated_maps.go
```diff
@@ -0,0 +1,55 @@
...
+				"thumbs": []
+			}
+		}`), new_image)
+		collection.Schema.AddField(new_image)
+
+		return dao.SaveCollection(collection)
+	}, func(db dbx.Builder) error {
+		dao := daos.New(db);
+
+		collection, err := dao.FindCollectionByNameOrId("tgcm0cd7l7jj0ze")
```
Flizzyclone, line 45: Should `FindCollectionByNameOrId("tgcm0cd7l7jj0ze")` and `collection.Schema.RemoveField("q3odadwo")` be hardcoded? They could change with each instance or yapms.

apps/pocketbase/migrations/1679451083_updated_maps.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/pocketbase/pocketbase/daos"
+	m "github.com/pocketbase/pocketbase/migrations"
+	"github.com/pocketbase/pocketbase/models/schema"
+)
+
+func init() {
+	m.Register(func(db dbx.Builder) error {
+		dao := daos.New(db);
+
+		collection, err := dao.FindCollectionByNameOrId("tgcm0cd7l7jj0ze")
```
Flizzyclone, line 16: See above comment about hardcoding

Flizzyclone (changes requested):

apps/pocketbase/main.go
Fish1: Yes, thanks.

apps/pocketbase/main.go
Fish1: Yeah, but then we can't control the size of the screenshot easily. Because the sidebar and topbar use up a certain amount of pixels. I'd rather add something to the frontend that renders the map as it should be screenshot.

apps/pocketbase/migrations/1679449550_updated_maps.go
Fish1: From my understanding, the ID should be constant.

I didn't write these migration files, pocketbase automatically creates them when you modify the database.

You can see the "collections_snapshot.go" file it will generate a table called maps with the "tgcm0cd7l7jj0ze" id. So anyone who runs pocketbase will have the same ID.

Then it will run through all the migration files to make the correct modifications to the maps database.

But these files aren't written by hand, its all auto generated and executed by pocketbase.

image

Fish1: This function enables automigrations.

image

Fish1: aaaa

Flizzyclone (approved):
LGTM!

PR #134, opened 2023-03-26
Remove pocketbase routes & use SDK

### LINKED ISSUE #132: Screenshots

Fish1:
Feature: Pocketbase has built in capabilities to add records to the database. So I think its better to use that instead of building our own custom routes and using the fetch API.

Code: Removed the route in pocketbase, and used the OnRecordBeforeCreate and OnRecordAfterCreate. These get called whenever a record gets added to the system.

I also removed the hash in the map collection, because I don't think it would actually prevent that much duplicate data. There are so many things yapms can customize, the likelihood of two people making the same map is probably pretty low.

closes #132

### DIFF
--- a/apps/pocketbase/main.go
+++ b/apps/pocketbase/main.go
@@ -4,23 +4,16 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
-	"crypto/md5"
-	"encoding/base64"
 	"flag"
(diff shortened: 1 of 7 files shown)


Flizzyclone (approved):
LGTM, good call on the hash

PR #136, opened 2023-03-28
Map Search

### LINKED ISSUE #120: Map Search
As we add more and more maps, we will need a way for users to search for them on the home page. The bar itself can go up here:
image

Flizzyclone:
Adds a search bar to the top of the home page based on Titles.json
yapmsSearch2

Summary of Code Changes:
Adds package svelte-typeahead to do fuzzy searching
Adds search bar component using Typeahead component
Adds CSS for the search bar to global.css, from there I use @apply to use tailwind styles (we can break this out into it's own file if we want, but we need to use CSS)
Changes padding for the divider on the home page so the search bar outline is not cut off when selected.

Closes #120


Flizzyclone:
Changed to using a unicode icon for the search magnifying glass- absolute positioning of svg works for daisyui but not for us. Typeahead does not have support for an icon within the input.

image

Fish1:
Can you elaborate on what the issue was with the SVG? It seemed to be working for me.

I am against the Unicode emojis because, they look different depending on the users operating system, and the don't necessarily fit the style or current theme of the website.
image

Flizzyclone:
Sure! Issue with the SVG was that it had to be positioned absolutely and that meant that whenever you scrolled the page the SVG would leave the search bar.

Now that I'm thinking about it, I might be able to solve it by breaking out the search bar from the scrollable div. I will test this in a bit.

Flizzyclone:
I did some work on the css and using a negative margin I got the svg icon to work as desired. Will recommit the SVG in a second here.

apps/yapms/src/lib/components/homepage/MapSearch.svelte
```diff
@@ -0,0 +1,37 @@
...
+<div class="pt-6">
+	<div class="px-6 md:px-0 w-full lg:w-4/6 flex">
+		<Search class={`pointer-events-none stroke-current my-3 pt-1 w-9 z-10 ml-2`}
+			><path
+				stroke-linecap="round"
+				stroke-linejoin="round"
+				stroke-width="2"
+				d="M21 21l-6-6m2-5a7 7 0 11-14 0 7 7 0 0114 0z"
+			/></Search
+		>
```
Fish1: What is this path in the Search component for?

apps/yapms/src/lib/icons/Search.svelte
```diff
@@ -0,0 +1,21 @@
...
+	stroke="currentColor"
+	class={className}
+	{style}
+>
+	<path
+		stroke-linecap="round"
+		stroke-linejoin="round"
+		d="M21 21l-5.197-5.197m0 0A7.5 7.5 0 105.196 5.196a7.5 7.5 0 0010.607 10.607z"
+	/>
+</svg>
```
Fish1: This file should be called MagnifyingGlass. Just a bit more descriptive so you know what it's going to look like while reading the code.

Fish1 (changes requested):
Looks good and seems to function well.

Just a name change for the SVG, and a question about the path.

apps/yapms/src/lib/components/homepage/MapSearch.svelte
Flizzyclone: Whoops thats a error from when I was copy pasting the svg. Removed

apps/yapms/src/lib/icons/Search.svelte
Flizzyclone: Changed!

Fish1 (approved):
LGTM

PR #142, opened 2023-04-04
Add USA Senate maps

Flizzyclone:
This PR adds 6 different variants of the USA Senate map:
1. Blank Senate
2. Current Senate
3. Senate 2024 Elections
4/5/6. Senate Class 1/Class 2/Class 3

A few notes about these maps:
1. I used different colors for Democrats and Republicans than the testing map uses. I think these look better but we can change them back if we want.
2. In order to use the Roboto font, which iirc we decided on as default font, I had to add a CSS import for the font, hence the addition of map.css.
3. In the GitHub previewer these maps will appear to have no text/color. This is to keep file size down. They look fine in app.

Code changes:
1. Updated map selection with the proper link to the Senate 2024 map
2. Added the maps described above
3. Added Senate maps to Titles.json
4. Created and added import for map.css in order to use the Roboto font.


apps/yapms/src/routes/app/+layout.svelte
```diff
@@ -1,5 +1,6 @@
 <script>
 	import '$lib/styles/global.css';
+	import '$lib/styles/map.css';
```
Fish1, line 3: I don't think we need this. The font should be built directly into the map I think. Removing it didn't change anything for me.

Fish1 (changes requested):

apps/yapms/src/routes/app/+layout.svelte
Flizzyclone: On Chrome when I remove this, the map reverts back to a Times New Roman style font.
image

Fish1: Ahh, maybe it's because I have Roboto installed directly on my operating system.

Fish1 (approved):
LGTM

PR #144, opened 2023-04-06
Added "None" Chart Type

Flizzyclone:
This PR adds a "None" chart type that hides the div that the chart resides in.

image
*When 'none' is selected.*

Summary of code changes:
- Adds 'none' as a possible value for the Chart Type Modal
- Adds 'none' as a possible value for the ChartType type
- Adds a statement to add the hidden class to the chart-containing div when ChartTypeStore is 'none'.


Fish1 (approved):
LGTM

PR #148, opened 2023-04-11
Default-mode property

### LINKED ISSUE #147: defaultMode property for maps
For some maps like presidential primary maps, it will be useful to have a "defaultMode" property that allows you to specify a mode other than fill to use when the map is opened.

Flizzyclone:
This PR adds a code to interpret a "default-mode" property on map SVG's on load. This property will be used to set the mode to something other than fill when a map requires it, e.g a presidential primary map where you would want to set it to "split" mode.

Summary of code changes:
- Adds logic in the LoadRegions function to set the mode based on the default-mode property
- Sets the default-mode property to "split" for the Danish Folketing map

Closes #147


apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -114,6 +115,12 @@ function loadRegions(node: HTMLDivElement): void {
 		console.error('Error Parsing Candidate Data from Map:\n\n' + error);
 	}
 
+	//If map being loaded has the default-mode property set, change the mode.
+	const defaultMode = node.querySelector('svg')?.getAttribute('default-mode');
+	if (defaultMode !== null) {
+		ModeStore.set(defaultMode as Mode);
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Here, you need to check for the undefined condition, and each possible string value of Mode. Currently the cast to Mode is causing undefined type to be hidden.

```
if (defaultMode !== null && defaultMode !== undefined && (defaultMode === "edit" ... ))
```
This will reduce the string type down to the Mode type.

AFAIK this is the only way... unless there is some fancy typescripting that can be done with the Mode type.

Fish1 (changes requested):
Type changes on defaultMode

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Ah rats.. will check for a way to maybe do this in zod tomorrow.

Comment (Fish1) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Oh yeah, I forgot about zod lol. Yeah, that's a better idea.

Flizzyclone:
Checking if defaultMode is valid is now handled by zod, and the Mode type is constructed from a zod enum. Please re-review.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -114,6 +116,12 @@ function loadRegions(node: HTMLDivElement): void {
 		console.error('Error Parsing Candidate Data from Map:\n\n' + error);
 	}
 
+	//If map being loaded has the default-mode property set, change the mode.
+	const defaultMode = node.querySelector('svg')?.getAttribute('default-mode');
+	if (ModeSchema.parse(defaultMode)) {
+		ModeStore.set(defaultMode as Mode);
+	}
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: With Zod we don't need to cast any data.

We can either use parse or safe parse, and use type narrowing.

parse will throw an error you can catch
```ts
try {
  const defaultModeAttribute = node.querySelector('svg')?.getAttribute('default-mode');
  const defaultMode = ModeSchema.parse(defaultModeAttribute);
  ModeStore.set(defaultMode);
} catch (error) {
  console.error('Error Parsing Mode Data from Map:\n\n' + error);
}
```

safeParse will return an object with with data about if it's an error or not.
```ts
const defaultModeAttribute = node.querySelector('svg')?.getAttribute('default-mode');
const defaultMode = ModeSchema.safeParse(defaultModeAttribute);
if (defaultMode.success) {
  ModeStore.set(defaultMode.data);
}
```

I think whichever one you think is cleanest will be fine here.

Comment (Fish1) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: By having the parse in the function, it will cause the loadRegions function to throw an error if it finds an invalid default mode. It should default to something if it doesn't load proper data.

Fish1 (changes requested):
Your solution does work, but you can use the type narrowing provided by Zod so you don't need to do any casting.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Right, good catch. Will change to safeParse.

Flizzyclone:
Changed to safeParse. Please re-review :)

Fish1 (approved):
Sorry, missed this.

LGTM

PR #150, opened 2023-05-06
USA House Maps 1992-2024

Flizzyclone:
This PR adds USA House Maps from 1992-2024. You can preview these at https://yapmspreview.flizzy.dev/.

Average file size is ~1500kb, this is in-line with the 2022 house map on yapms1.

Sourcing:
1992-2012 are made from shapefiles from https://cdmaps.polisci.ucla.edu/
2014-2018 are made from shapefiles from the [U.S Census bureau](https://www.census.gov/geographies/mapping-files/time-series/geo/carto-boundary-file.html)
[2020](https://en.wikipedia.org/wiki/2020_United_States_House_of_Representatives_elections), [2022 ](https://en.wikipedia.org/wiki/2022_United_States_House_of_Representatives_elections) and therefore 2024 are adapted versions of Wikipedia maps.

These maps switched to using a new color scheme. We should probably apply this onto other maps.


Fish1 (approved):
LGTM

PR #155, opened 2023-05-20
Default candidate counts support

### LINKED ISSUE #154: Starting count for candidates
We need a property either within the candidates argument or as an additional property on the svg that allows starting seats to be given to candidates

Flizzyclone:
This PR adds a code to interpret a "candidate-counts" property on map SVG's on load. This property is used to set the "default" candidate counts. This is useful for maps where not all seats are up (US senate).

This property is somewhat confusing at first glance but it really is just an array of objects containing a single key value pair (an array of objects seems to be better for zod type checking than a undefined length list of key value pairs). This key value pair is structured like the following:
`"candidate id (str)": default count (num)`
so, an example property may look like:
`candidate-counts=[{"0":12},{"1":26}]`

Summary of code changes:
- Creates DefaultCountsStore for storing the counts in a Map that can be used in Regions.ts when deriving other stores. Default value for this store is an empty map
- Creates DefaultCountsSchema for checking that data passed in the prop is correct
- Adds logic in LoadRegions.ts to interpret this property and set DefaultCountsStore to its proper value
- Changes the starting map when deriving CandidateCounts to DefaultCountsStore
- Changes the starting map when deriving CandidateCountsMargins to be a version of DefaultCountsStore adapted to the format of CandidateCountsMargins
- Adds default counts to the 2024 Senate map

Closes #154


apps/yapms/src/lib/assets/maps/usa-senate-2024.svg
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg candidates='[{"id":"0", "name":"Democrats", "margins": [{ "color": "#1C408C" }, { "color": "#577CCC" }, { "color": "#8AAFFF" }, { "color": "#949BB3" }]},    {"id":"1", "name":"Republicans", "margins": [{ "color": "#BF1D29" }, { "color": "#FF5865" }, { "color": "#FF8B98" }, { "color": "#CF8980" }]}]' version="1.1" viewBox="0 0 959 593" xmlns="http://www.w3.org/2000/svg">
+<svg candidate-counts='[{"0":28},{"1":38}]' candidates='[{"id":"0", "name":"Democrats", "margins": [{ "color": "#1C408C" }, { "color": "#577CCC" }, { "color": "#8AAFFF" }, { "color": "#949BB3" }]},    {"id":"1", "name":"Republicans", "margins": [{ "color": "#BF1D29" }, { "color": "#FF5865" }, { "color": "#FF8B98" }, { "color": "#CF8980" }]}]' version="1.1" viewBox="0 0 959 593" xmlns="http://www.w3.org/2000/svg">
```
Fish1: Would it be possible to instead have the starting count put inside the candidates property?

```[{"id": "0", "name":"Democrats", "count":"38"}]```

**Review by Fish1 — CHANGES_REQUESTED** [review]
Could we put the data from candidate-counts inside candidates?

File: apps/yapms/src/lib/assets/maps/usa-senate-2024.svg (hunk 1 above)
Comment (Flizzyclone) [reply apps/yapms/src/lib/assets/maps/usa-senate-2024.svg]: Yep! Just pushed this. Less complex than I thought.

I changed the following
- Removed DefaultCountsSchema and DefaultCountsStoreSchema as all checking is now done w/ CandidateSchema
- Added an optional counts number argument to CandidateSchema
- Changed the Senate map to follow new formatting
- Added a null check for default-mode since it kept throwing errors

**Comment by Flizzyclone** [comment]
Alright this PR now adds a "permalocked" property that removes the region from play and is irreversible by the user. No functionality is given to these permalocked states.

Changes:
- Adds "permalocked" property to Region class and the logic to set this on map load.
- Changes some display logic to make sure permalocked states appear without text and faded
- Changes states on 2024 Senate to permalocked instead of simply disabled:
image

File: apps/yapms/src/lib/types/Region.ts [hunk 2]
```diff
@@ -9,6 +9,7 @@ export const RegionSchema = z.object({
 	permaVal: z.number(),
 	disabled: z.boolean(),
 	locked: z.boolean(),
+	permalocked: z.boolean(),
```
Fish1: permalocked should be permaLocked just to match the camel case pattern.

apps/yapms/src/lib/stores/Candidates.ts
```diff
@@ -33,6 +33,8 @@ export const CandidatesStore = writable<Candidate[]>([
 	}
 ]);
 
+export const DefaultCountsStore = writable<Map<string, number>>(new Map<string, number>());
```
Fish1: The DefaultCountsStore is redundant data, as you've already done all the work to collect this information into the CandidateStore.

apps/yapms/src/lib/types/Candidate.ts
```diff
@@ -3,6 +3,7 @@ import { z } from 'zod';
 export const CandidateSchema = z.object({
 	id: z.string(),
 	name: z.string(),
+	count: z.number().optional(),
```
Fish1: This property already contains the data for the default count. Maybe this could be named something else such as "defaultCount" or "permaCount" or something that represents it's not just the current count.

Maybe this could also be non-optional and just have it defaulted to zero. That would make the calculations for where this is used a bit easier.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -111,17 +112,26 @@ function loadRegions(node: HTMLDivElement): void {
...
-		ModeStore.set(defaultMode.data);
-	} else {
-		console.error('Error Parsing defaultMode attribute from Map:\n\n' + defaultMode.error);
+	if (defaultModeAttribute !== null) {
+		const defaultMode = ModeSchema.safeParse(defaultModeAttribute);
+		if (defaultMode.success) {
+			ModeStore.set(defaultMode.data);
+		} else {
+			console.error('Error Parsing defaultMode attribute from Map:\n\n' + defaultMode.error);
+		}
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: I think all of the attributes on the maps should be 100% required. That way when a map is created, it's 100% consistent and doesn't rely on the default behaviors of the website.

If we ever change the default behavior of the website, we need to look for where that default behavior change doesn't make sense, which means checking every single map without an explicit parameter. Where as if it is explicit, we have fewer situations like that.

Fish1 (changes requested):
The DefaultCountsStore has it's data duplicated in the CandidateStore itself, we can just use the candidate store.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: I will break out making all attributes required into a separate Issue/PR and leave it untouched in this one.

Comment (Fish1) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Ahh Looking at the code, I guess it was never required in the first place. My bad.

But I think the original behavior of logging the error on null was fine here.

Flizzyclone:
Alright all issues should be resolved here-
- I renamed the count property to "defaultCount" and made it required w/ a default of 0. Some logic was updated to reflect this change
- I removed DefaultCountsStore and modified the CandidateCounts logic in Regions.ts to use CandidatesStore.
- permalocked is now permaLocked
- I changed some logic in AddCandidates and EditCandidates to account for the additional property

Fish1 (approved):
LGTM!

PR #156, opened 2023-05-20
pull out custom pocketbase logic into their own files

Fish1:
Code: pulled out some pocketbase logic into their own files. Should make the mainline code more readable.

Clearly shows that before a record is created, we compress the map file.

After a record is created we then use browserless to take a screenshot.

image


Flizzyclone (approved):
Nice, this is much better.

PR #163, opened 2023-05-26
Screenshot View

### LINKED ISSUE #161: Better Screenshots
Browserless simply takes a screenshot of the entire page.

image

I think having a page dedicated to displaying the map for a screenshot would be best. Something like...

/app/view
/app/screenshot
/app/display

Fish1:
Features: Added a /view route that will load and render the map as fast as possible. It will not load unneeded things such as popups, the topbar or sidebar.

Code: 

1) The candidate box now takes in a "selectable" parameter. If it is false, it will default to being small.

2) The battle chart can now handle multiple candidates. It just displays all their margins in a row, and doesn't have the circle with an arrow icon.

3) Refactored much of the region logic out of the map initialization code. The function specific to modifying a region (such as filling in, editing and disabling) is now in the /stores/regions directory.

4) There are now two loadRegions function. One for the app and one for the view. Loading the regions for the view simply does less work.
image

Example of the /view route.
http://localhost:8081/view?m=vasx1atwb4u9bqw
image

closes #161


apps/yapms/src/lib/stores/regions/Regions.ts
```diff
@@ -143,3 +146,46 @@ export const CandidateCountsMargins = derived(RegionsStore, ($RegionStore) => {
 	});
 	return candidates;
 });
+
+export const setPointerEvents = (): void => {
```
Flizzyclone, line 150: Is there a reason that this function specifically is not with the rest called on initialization?

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -179,61 +106,36 @@ function loadRegions(node: HTMLDivElement): void {
...
+	setDefaultMode(node);
+	setCursorStyle(node);
+	createRegionStore(node);
+	setTransitionStyle(node);
+	setPointerEvents();
+}
+
+export function loadRegionsForView(node: HTMLDivElement): void {
+	loadCandidateData(node);
+	setDefaultMode(node);
```
Comment (Flizzyclone) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: I don't think we need setDefaultMode here since there is no editing in the view endpoint

Flizzyclone (changes requested):
One more thing I noticed is that when I load the view page, all fonts are in Times New Roman as I don't have Roboto installed on my system. This  is fixed by adding a +layout.svelte for that endpoint containing `import '$lib/styles/map.css';`. I don't know if this would be a problem for screenshots or not.

apps/yapms/src/lib/stores/regions/Regions.ts
Fish1: Good point, I think just because it doesn't parse the node data. It was just reading the regions.

I think I'll make it so that more function use the region store instead of the raw nodes.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Fish1) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Fixed

Fish1:
I imported the fonts, moved all the functions to the region folder, and removed the default mode function from the view.

Flizzyclone (approved):
LGTM then!

PR #165, opened 2023-05-30
Standard colors for Pres 2022

Flizzyclone:
This PR changes the candidates for the presidential 2022 map in order to put margin colors in line with the rest of the USA maps.


Fish1 (approved):
LGTM

PR #169, opened 2023-06-07
Imports from Subdirectories

### LINKED ISSUE #166: Subdirectory map imports
As we add more and more maps, the map asset directory is becoming increasingly unwieldy. To address this, I plan on updating the map import code to take in maps contained in subdirectories of the map directory.

Flizzyclone:
This PR changes import logic to import from subdirectories instead of the main maps directory, allowing us to better organize map files.

Summary of code changes
- Updates import logic in both the /app and /view routes to import from all subdirectories 
- Adds logic while importing to strip information about subdirectories from the keys of import object so the exact directory does not need to be known in order to import, only the map name
- Moves all existing maps into subdirectories

Closes #166


Fish1 (approved):
LGTM

PR #173, opened 2023-06-10
Beta Modal & Removal of Nonexistent Links

### LINKED ISSUE #170: Beta warning popup
The yapms2.com domain is indexed by google. People should know they are using a beta product.

### LINKED ISSUE #172: Non-2024 Maps
Any attempt to open a map besides US Pres, Senate, and House opens up US Pres 2024 map

Flizzyclone:
This PR adds a modal containing a notice that yapms2 is in beta to the home page. This modal will only show up in prod, so to see it you will need to run `npm run build` and then `npm run preview`. I do this by using `!dev` as a condition for the modal to show up.

image

This PR also removes cards on the home page for nations other than the U.S, as those links would lead back to the USA 2024 map and were causing confusion.

Summary of Code Changes:
- Adds BetaModal & BetaModalStore
- Removes countries other than USA from MapSelection.json

Closes #172 and #170

### DIFF
--- a/apps/yapms/src/lib/components/modals/betamodal/BetaModal.svelte
+++ b/apps/yapms/src/lib/components/modals/betamodal/BetaModal.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-	import ModalTitle from '$lib/components/modalutilities/ModalTitle.svelte';
 	import { BetaModalStore } from '$lib/stores/Modals';
 	import { dev } from '$app/environment';
 	import ExclamationCircle from '$lib/icons/ExclamationCircle.svelte';
--- a/apps/yapms/src/lib/components/modals/betamodal/BetaModal.svelte
+++ b/apps/yapms/src/lib/components/modals/betamodal/BetaModal.svelte
@@ -1,0 +1,50 @@
+<script lang="ts">
+	import ModalTitle from '$lib/components/modalutilities/ModalTitle.svelte';
+	import { BetaModalStore } from '$lib/stores/Modals';
+	import { dev } from '$app/environment';
+	import ExclamationCircle from '$lib/icons/ExclamationCircle.svelte';
+	import Fa from 'svelte-fa';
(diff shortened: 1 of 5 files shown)


Flizzyclone:
Ope sorry Fish, didn't see that you assigned yourself to this one. Feel free to rewrite as needed.

Fish1 (changes requested):
It looks good!

If you could just use the ModalBase layout, that would be good. I've been using it for new modals, just haven't gone through and replaced them all. It just removes some boiler plate, and lets us tweak the layout of the modals without needing to change all the files.

The edittossupmodal uses it like this.
image


Also, if you could use the link component on daisyui, that way the colors will adjust with the theme.
https://daisyui.com/components/link/

As for the exclamation point... I know in react you can pass JSX as a variable. I don't think I've tried it with svelte, but maybe you can add a "icon" variable to the ModalBase component?

apps/yapms/src/lib/components/modals/betamodal/BetaModal.svelte
```diff
@@ -0,0 +1,50 @@
...
+		<p />
+		<br />
+		<p>
+			You can report bugs, suggest features, and contribute to development on our <a
+				class="text-blue-500 underline"
+				href="https://github.com/yapms/yapms">GitHub</a
+			>
+			or in our <a class="text-blue-500 underline" href="https://discord.gg/Rq5bk3eDwm">Discord</a>.
+			Thank you.
+		</p>
```
Fish1: This layout should be controlled by a flex box with a gap, like the above title is. But the flex box can flex vertically.

Fish1 (changes requested):

Flizzyclone:
I have addressed all issues. For the icon, I added an additional slot to ModalBase since you cannot pass HTML like you can with JSX in React. This doesn't throw off other modals using ModalBase, just allows us to add in icons.

Fish1 (approved):
Oh yeah, I like the slot idea, nice!

LGTM

PR #174, opened 2023-06-12
Umami Analytics

### LINKED ISSUE #69: Umami Analytics
Umami is an open source alternative to google analytics.

Seem to be a bit cleaner of an interface to use, and we just need simple usage data, not everything provided by google.

https://umami.is/

Fish1:
Feature: Host umami analytics locally on the server. We can generate a link to give everyone access to the analytics.

Code: Added mariadb and umami images to the docker-development file, and the deploy.yml. The URL for umami is built into the frontend from the build process. We can use a volume for mariadb so the data doesn't get deleted on deployment.

image

closes #69


Fish1:
I'm not sure what is up with the github action... But I think the check step is working...

image

Flizzyclone:
Re: the GitHub Action, it looks like the .env that it pulls for the check doesn't have the new environment variables you defined.

Flizzyclone (approved):
LGTM!

PR #176, opened 2023-06-12
Pie chart max size & Chart Options modal

Flizzyclone:
This PR addresses the bottom pie chart being [huge](https://media.discordapp.net/attachments/1038875719666765954/1117511207058948126/Captura_de_tela_2023-06-11_144124.png?width=1038&height=671) as well as putting chart type & position options into the same modal.

New pie chart sizing looks like this:
image

image

image

Chart options modal:
image

Summary of Code Changes:
- Adds a max height class to the pie chart div to control sizing
- Changes Chart Type modal into ChartOptionsModal
- Deletes Chart Position options from StyleModal
- Changes ChartPositionStore to be constructed from the type rather than an old hardcoded list.


Fish1 (approved):
LGTM, I've just removed the toUpperCase function call.

PR #177, opened 2023-06-12
Close sidebar

### LINKED ISSUE #171: Side bar minimize
The side bar should have a button to close/open it.

Fish1:
Feature: Added a button at the top-right of the navbar, it will close the sidebar.

Code: Created a sidebar store that stores a boolean. Use that boolean to conditionally render the sidebar.

Animation

closes #171

### DIFF
--- a/apps/yapms/src/lib/components/sidebar/SideBar.svelte
+++ b/apps/yapms/src/lib/components/sidebar/SideBar.svelte
@@ -5,23 +5,25 @@
 	import { faReddit, faTwitter } from '@fortawesome/free-brands-svg-icons';
 
 	import titles from '$lib/assets/other/Titles.json';
+	import { SideBarStore } from '$lib/stores/SideBar';
 
 	$: title = titles.find((elem) => elem.path === $page.url.pathname)?.title ?? 'YAPms';
 </script>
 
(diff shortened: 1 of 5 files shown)


apps/yapms/src/lib/components/sidebar/SideBar.svelte
```diff
@@ -5,23 +5,25 @@
...
-		<button type="button" class="btn btn-sm gap-2">
-			<Fa icon={faReddit} />
-			Reddit
-		</button>
-		<button type="button" class="btn btn-sm gap-2">
-			<Fa icon={faTwitter} />
-			Twitter
-		</button>
+{#if $SideBarStore}
+	<div class="divider divider-horizontal w-0 m-0 hidden md:flex" />
```
Flizzyclone, line 14: `hidden md:flex` here means that the side bar can't be opened on mobile, making the button confusing. The way I see it, the best way to do that right now is to hide the button on mobile until we can rework the side bar to work on mobile.

Flizzyclone (changes requested):

apps/yapms/src/lib/components/sidebar/SideBar.svelte
Fish1: Added the same properties on the button, so now it will be hidden.

PR #181, opened 2023-06-15
Auto Stroke Borders

Flizzyclone:
This PR adds support for auto resizing stroke borders that match the background color and resize on zoom. This allows for finer detail in areas where districts are dense, like urban areas. You can test this by navigating to /app/debug/auto/borders

Offset through Inkscape:
image

Auto borders:
image

The downside is that applying these types of borders causes lag when you have many SVG objects on screen. When you zoom in, this isn't an issue. I would only recommend this for maps with many districts with potential to be densely packed, like a US house or Canadian House of Commons map.

Summary of code changes:
- Adds a CSS variable (auto-stroke-width) to any svg with the autoBorders class and uses that to add strokes to region paths
- Adds logic to adjust auto-stroke-width and therefore border as you zoom

Thanks to @a2435191 for the idea (get a better username, Will)


Flizzyclone:
All house maps updated to use this border style. Please review

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
```diff
@@ -3,13 +3,33 @@ import panzoom, { type PanZoom } from 'panzoom';
...
+	}
+}
+
+function adjustStroke(svg: SVGElement, scale: number) {
+	const initStroke = svg.hasAttribute('auto-border-stroke')
+		? Number(svg.getAttribute('auto-border-stroke'))
+		: 0.5;
+	const strokeUpper = Number(svg.getAttribute('auto-border-upper'));
+	const newStroke = initStroke / scale;
+	if (strokeUpper && newStroke > strokeUpper) {
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: If strokeUpper is 0 it will allow for unlimited stroke width.

Fish1 (changes requested):
Looks amazing! Love it!

I think the requirement for adding autoBorders to the map can be avoided. Instead of targeting .autoBorders, you can target the div that contains the SVG, and then the SVG itself. This way people making maps only need to set the --auto-border properties and don't need the class name.

image

---

I think the naming of auto-border-stroke and auto-border-upper can be improved. In SVG the stroke property refers to the color, so maybe "auto-border-stroke-width" and "auto-border-stroke-width-limit" would make it match the SVG conventions.

---

Nit pick: Pulling out the svg attribute functions, can help clean up the logic of the function.
image

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: This is fine-- just remove the attributes in that case for no stroke

Flizzyclone:
Alright I have addressed the above changes:
- I removed the requirement for auto-borders class and added an ID to the div the map gets loaded into so I can target it specifically instead of all svg's in all div's.
- Renamed attributes to "auto-border-stroke-width" and "auto-border-stroke-width-limit"
- Moved attributes out of function logic
```
#map-div > svg {
	--auto-border-stroke-width: 0px;
}
```

This bit of code is a bit confusing. I set the var to 0 because if not it will default to 1px and a 1px stroke will be applied to every map regardless of whether they have the stroke-width attribute or not. There does not seem to be any additional lag associated w/ having this variable on all maps as long as it is set to 0.

Please re-review.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
Comment (Fish1) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: But if someone sets the upper limit to zero, the software should cap the limit at zero. From an end user's perspective, when they are trying to make a map, they won't understand why setting the upper limit to zero results in unlimited upper limit (without reading the source code).

apps/yapms/src/lib/styles/global.css
```diff
@@ -25,3 +25,12 @@ body {
...
 }
+
+#map-div > svg {
+	--auto-border-stroke-width: 0px;
+}
+
+svg > .regions > path {
+	stroke-width: var(--auto-border-stroke-width);
+	stroke: hsla(var(--b1) / var(--tw-bg-opacity, 1));
+}
```
Fish1: Should this also only target the #map-div? I'll do a real review with testing when I get home tonight, just wanted to point it out.

Flizzyclone: Ahh yes good point

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: In the case that no upper limit is set the code will return NaN now (instead of the Number(undefined) == 0) and won't pass the check. 0 will now result in no stroke

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
```diff
@@ -3,13 +3,36 @@ import panzoom, { type PanZoom } from 'panzoom';
...
 			smoothScroll: false,
 			autocenter: true,
 			zoomDoubleClickSpeed: 1
 		});
+		if (svg.hasAttribute('auto-border-stroke-width')) {
+			const initStroke = Number(svg.getAttribute('auto-border-stroke-width'));
+			const strokeUpper = svg.hasAttribute('auto-border-stroke-width-limit')
+				? Number(svg.getAttribute('auto-border-stroke-width-limit'))
+				: NaN;
+			adjustStroke(svg, initStroke, strokeUpper, panzoomInstance.getTransform().scale);
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: This looks good, if you could just change the NaN to a null, and check for null instead I think that will be good. Generally NaN is reserved for errors in math computations, and not for control logic.


---

Nit: If you wanna handle all possible errors from user inputs, I think zod would be good here.

Users can pass in gibberish to generate NaN or "Infinity" to generate infinity.

By using something like z.number().finite().positive() you can verify that the user enters correct numbers.

image

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
```diff
@@ -3,13 +3,36 @@ import panzoom, { type PanZoom } from 'panzoom';
...
+			panzoomInstance.on('zoom', (e: PanZoom) => {
+				const initStroke = Number(svg.getAttribute('auto-border-stroke-width'));
+				const strokeUpper = Number(svg.getAttribute('auto-border-stroke-width-limit'));
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: Nit: You can use the initStroke and strokeUpper that is defined before this.

Fish1 (changes requested):
Looks good, if you could just change NaN usage to null.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: Oh I like this, will implement in a way that if initstroke is invalid, will return early.

Flizzyclone:
Switched to zod & null!

Fish1 (approved):
LGTM

PR #185, opened 2023-06-20
File importing

### LINKED ISSUE #184: tossup-candidate property
We need a property to specify the tossup color/name when loading in a map. This is needed for when we export an imported file. I will do this in the file importing PR.

Flizzyclone:
This PR adds importing into yapms from GeoJSON files, .shp files, and opening a YAPms SVG without having to go through cloning the git. There is a live build of this at https://yapmspreview.flizzy.dev/

This functionality uses a couple new dependencies. d3 is used to do the bulk of the work, taking in GeoJSON and outputting an SVG in Mercator projection. shapefile turns .shp files into GeoJSON that can then be run through d3, @turf/rewind orders unordered GeoJSON elements so d3 can read the GeoJSON without outputting a garbled mess.

Summary of Code Changes:
__UI:__
- Adds a new route, /app/imported for all imported maps. If no map is yet imported, this will come up with an import modal that allows you to select files. Once a map is imported, it will change to the normal app interface.
- Adds ImportModal for selecting files to be imported
- Adds ShareDisabledModal, which shows a notice that maps cannot be shared by link and offers the option to export as a yapms formatted SVG for sharing
- Adds some conditional rendering to the navbar for imported map specific buttons.

__Candidate Changes:__
- Adds the tossup-candidate attribute to SVG. Works like the candidates attribute, just not an array.
- Adds the ImportedSVG store to store the importedSVG & load status
- Makes TossupCandidateStore and CandidatesStore derived from also-exported variables containing their default values.
- Adds logic to reset Candidates to default values if the map being loaded does not specify candidates or tossup-candidate attributes
- Fixes an issue where if candidate ID's did not match their position in the candidate array they would be loaded incorrectly

__Import/Export Functionality:__
- Adds functions to turn .shp files and GeoJSON into a standardized format (importFromShapefiles, importFromGeoJson)
- Adds a function to change GeoJSON to SVG and store that ImportedSVG store (geoJsonToSVG)
- Adds a function to clear imported data to make way for a new map (newImportedMap)
- Adds a function to export the currently imported map as yapms formatted SVG (exportImportAsSVG)

Closes #184


Fish1 (commented):
I'll hold off on reviewing the entire thing.

With the current implementation of this, it requires a duplication of the /app/country/map/year route into the /app/imported route. So any code on one of the pages needs to be implemented twice for both pages.

Because the layouts are identical, I think the commonalities should be pulled down into the /app/+layout.svelte file. (Chart layouts, sidebar, navbar, dividers, candidate box, modals etc...)

Let me know what you think.

Fish1:
If you want, I could review the functionality of the code, and merge it in. But we would need another issue to refactor it into a common layout. And then either me or you can work on it.

Or we can fix fix the layout in this PR.

Whichever you think would be easier.

Flizzyclone:
I thought about doing this and I agree but reckoned that doing such a big refactor while we were working on other features would've been a bad idea that made a bunch of merge conflicts. I think doing this on a separate branch would be good since I don't know if the main app route has been touched in the past two weeks and the imported route hasn't.

Fish1:
Could you do a rebase onto main?

On the file imports you've added "--imported-stroke-width" but I am not sure how that fits in with the "auto-border-stroke-width" feature you've added.

image

PR #186, opened 2023-06-22
File importing

### LINKED ISSUE #184: tossup-candidate property

Flizzyclone:
This PR adds importing into yapms from GeoJSON files, .shp files, and opening a YAPms SVG without having to go through cloning the git. There is a live build of this at https://yapmspreview.flizzy.dev/

This functionality uses a couple new dependencies. d3 is used to do the bulk of the work, taking in GeoJSON and outputting an SVG in Mercator projection. shapefile turns .shp files into GeoJSON that can then be run through d3, @turf/rewind orders unordered GeoJSON elements so d3 can read the GeoJSON without outputting a garbled mess.

Summary of Code Changes:
__UI:__
- Adds a new route, /app/imported for all imported maps. If no map is yet imported, this will come up with an import modal that allows you to select files. Once a map is imported, it will change to the normal app interface.
- Adds ImportModal for selecting files to be imported
- Adds ShareDisabledModal, which shows a notice that maps cannot be shared by link and offers the option to export as a yapms formatted SVG for sharing
- Adds some conditional rendering to the navbar for imported map specific buttons.

__Candidate Changes:__
- Adds the tossup-candidate attribute to SVG. Works like the candidates attribute, just not an array.
- Adds the ImportedSVG store to store the importedSVG & load status
- Makes TossupCandidateStore and CandidatesStore derived from also-exported variables containing their default values.
- Adds logic to reset Candidates to default values if the map being loaded does not specify candidates or tossup-candidate attributes
- Fixes an issue where if candidate ID's did not match their position in the candidate array they would be loaded incorrectly

__Import/Export Functionality:__
- Adds functions to turn .shp files and GeoJSON into a standardized format (importFromShapefiles, importFromGeoJson)
- Adds a function to change GeoJSON to SVG and store that ImportedSVG store (geoJsonToSVG)
- Adds a function to clear imported data to make way for a new map (newImportedMap)
- Adds a function to export the currently imported map as yapms formatted SVG (exportImportAsSVG)

Closes #184


apps/yapms/src/lib/utils/importMap.ts
```diff
@@ -0,0 +1,107 @@
+import { ImportedSVGStore } from '$lib/stores/ImportedSVG';
+import { ImportModalStore } from '$lib/stores/Modals';
+import rewind from '@turf/rewind';
```
Fish1, line 3: Use DomPurify to cleanup XSS from imported SVGs.

https://github.com/cure53/DOMPurify

apps/yapms/src/lib/utils/importMap.ts
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	geoJsonToSVG(districtShapes);
+}
+
+function geoJsonToSVG(districtShapes: GeoJSON.FeatureCollection) {
+	const width = 1000,
+		height = 1000;
+
+	districtShapes.features = districtShapes.features.filter((feature) => feature.geometry != null);
```
Fish1: Use !== to avoid any type juggling.

apps/yapms/src/lib/utils/importMap.ts
```diff
@@ -0,0 +1,107 @@
...
+			`<path region="${i}" short-name="${i}" long-name="District ${i}" value="1" d="${render(
+				feature
+			)}"/>`
+	);
+
+	const output = `<svg id="importedSVG" width="${width}" height="${height}" auto-border-stroke-width="1" auto-border-stroke-width-limit=".1">
+    <g class="regions">
+    ${paths.join('\n')}
+    </g>
+    </svg>`;
```
Fish1, line 95: Probably should use createElements here and change their properties via the API.

apps/yapms/src/lib/components/modals/sharedisabledmodal/ShareDisabledModal.svelte
```diff
@@ -0,0 +1,35 @@
...
+	<div slot="icon"><ExclamationCircle class="w-9 h-9 text-red-500" /></div>
+	<div slot="content">
+		<div class="flex flex-col gap-y-2">
+			<p>
+				Due to <a
+					class="link link-primary"
+					href="https://rietta.com/blog/svg-xss-injection-attacks/">security concerns</a
+				> pertaining to storing and serving user generated SVG, sharing imported maps via link is not
+				supported. You can, however, export this map as a YAPms formatted SVG, which can be opened by
+				others. Only open or import files from people you trust!
```
Fish1: This can probably be shortened up, and we should be using something like DomPurify to remove XSS vulnerabilities.

Also, the reason that I'm not hosting svgs, is due to storage space.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -63,7 +85,9 @@ function createRegionStore(node: HTMLDivElement) {
 						//If the region has candidate-id defined, set the candidate appropriately, if not, default to the tossup candidate at margin 0
 						//See predetermined_regions_example.svg for an example of how to implement candidate-id and candidate-margin attributes.
 						{
-							candidate: get(CandidatesStore)[Number(childHTML.getAttribute('candidate-id'))],
+							candidate: get(CandidatesStore).filter(
+								(candidate) => candidate.id === childHTML.getAttribute('candidate-id')
+							)[0],
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Array.find would probably me more suitable here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

apps/yapms/src/routes/app/imported/+page.svelte
```diff
@@ -0,0 +1,137 @@
+<script lang="ts">
```
Fish1, line 1: Much of this page is a duplicate of the other page. We should move the layout out a common route.

apps/yapms/src/lib/components/modals/importmodal/ImportModal.svelte
```diff
@@ -0,0 +1,167 @@
+<script lang="ts">
```
Fish1, line 1: I find it a bit awkward to have this modal be the only thing on the /app/imported route with a blank background. Maybe this logic can be integrated directly into the homepage. And this modal can be used in the /app/country/map/year route, and cause a navigate to the /app/imported page.

Fish1 (approved):
LGTM!

Let me know if you wanna address any of my comments, otherwise I'll merge it in.

apps/yapms/src/lib/utils/importMap.ts
Flizzyclone: I will look into this but it may be a big refactor.

apps/yapms/src/routes/app/imported/+page.svelte
Flizzyclone: Putting this in a separate PR

Flizzyclone:
Alright, I addressed most of these requests.
- I reworked how the import modal works so it will show up over the existing page and redirect to /app/imported. 
  - There is now a button to open the import modal in /app/country/[country]/[map]/[year]
  - There is a failsafe in place for if you end up on /app/imported without any map loaded
- I added sanitization for both importing & exporting.
  - Whenever we add an attribute we need to add it in DOMPurifyConfig 
- I updated the ShareDisabledModal to be shorter and reflect the actual reason for the limited functionality.
- Changed that Array.filter() to a method that uses Array.find().

Please re-review as code has changed significantly.

Fish1 (approved):
LGTM

PR #202, opened 2023-06-28
MoreMapsModal Rework

### LINKED ISSUE #141: MoreMapsModal rework
Currently the MoreMapsModal reads from a json file called MapList.json. Now that we are using Titles.json for map search, I am going to update the MoreMapsModal to use Titles.json as well filtering based on path.

Flizzyclone:
This PR reworks MoreMapsModal by changing how it gets it's data from a json file to filtering Titles.json

Summary of Code Changes:
- Changes the route property on HomeLinkData to optional, adds optional routes prop
- Adds title prop to MoreMapsModalStore, changes key (string) prop to keys (string[])
- Changes the $maps variable in MoreMapsModal to being an array of Title objects matching paths defined in the routes property in MapSelection
- Adds routes for USA and U.S House

Closes #141


apps/yapms/src/lib/components/homepage/MapSelectionCard.svelte
```diff
@@ -10,11 +10,13 @@
...
 
-	function openMoreModal(key: string) {
+	function openMoreModal(keys: string[]) {
 		MoreMapsModalStore.set({
-			key,
+			keys,
+			title: name,
 			open: true
 		});
+		console.log(keys);
```
Fish1: Remove console logs.

apps/yapms/src/lib/types/HomeLinkData.ts
```diff
@@ -1,7 +1,8 @@
 type HomeLinkData = {
 	label: string;
 	modal: boolean;
-	route: string;
+	route?: string;
+	routes?: Array<string>;
 };
```
Fish1: It's a bit confusing to have this data struct serve dual purposes. The routes and route parameters are doing two things, but they can both be entered, I'm not sure what it would mean if a HomeLinkData had a route and a routes.

There should be two data structures, one that models clicking a button to open a map. And another that is for opening a modal.

apps/yapms/src/lib/components/modals/moremapsmodal/MoreMapsModal.svelte
```diff
@@ -1,33 +1,31 @@
...
 	}
-	//When the key updates, get the map data corresponding to that new key as defined in MapList.json. If none exists, empty data.
-	$: maps = mapList.find((elem) => elem.key === $MoreMapsModalStore.key) || {
-		label: '',
-		links: []
+
+	$: maps = titles.filter((title) => $MoreMapsModalStore.keys.indexOf(title.path) !== -1) || {
+		path: '',
+		title: ''
 	};
```
Fish1: I don't think the second expression of the or statement ever executes here. An array will always be true.

Fish1 (changes requested):

Flizzyclone:
Alright I have broken these into two data structures, HomeLinkData and HomeModalData. They will now be defined separately in MapSelection.json under the "links" and "modal" properties respectively. Modal buttons will always render after other links. Please re-review.

Fish1 (approved):
LGTM. Just for future note, when making type definitions, you don't need to add the word "Data" to the end of them.

PR #203, opened 2023-07-02
Proper selection of SVG when exporting

Flizzyclone:
This PR fixes an issue where when importing an SVG with an ID other than "importedSVG", you would be unable to export the map as the code was not looking into the map-div like we do elsewhere. 

The only code change here is the SVG being exported will now be found based on being a child of map-div instead of the SVG's ID.


Fish1 (approved):
LGTM

Issue #199, opened 2023-06-27
Show insets toggle


Flizzyclone:
For maps with insets, there should be an option to toggle whether or not the insets are shown. Probably in the style modal. We will probably need to define a class for inset elements that can be targeted by css to become hidden when a class is added to the svg itself.

PR #209, opened 2023-07-03
Insets

### LINKED ISSUE #46: Odd map behavior when changing chart position.
When chart in side position, the map doesn't fill it's entire space and is relegated to the top middle.
image

Removal of this transform tag applied to the svg element itself fixes it but results in unintended behavior for the bottom chart position:
image

image

### LINKED ISSUE #207: Inset Support
On some maps, insets are a nice to have. We currently don't have any proper support built in for them. To add support I plan to do the following:
- Add CSS classes for inset lines and text that set fill/stroke to background contrast color
- Add a CSS class for insets so they can be hidden when in a group with the uninsetted region.
- Add a toggleable class (toggle in style modal) to disable visibility of insets for those who don't want them.

Flizzyclone:
This PR adds support for insets. There is a debug map added to test these at /app/debug/inset/regions

The structure of an inset region looks like this (similar to how we do buttons for senate maps, but the inset has to have the inset-region class in order to be hidden):
image

I fixed the problem of the map not being properly centered when insets are hidden by temporarily setting SVG width and height attributes to the bounding box widths and heights for regions. Panzoom then centers based on this width, and then it is changed back so if you unhide insets they will display. Due to the way viewBox works, this will only work if width and height are used instead of viewBox.

Summary of Code Changes:
- Adds CSS rules to maintain background contrast on inset text & lines
- Adds CSS rules to hide inset regions when map SVG has the insetsHidden class
- Adds a checkbox to StyleModal to toggle the insetsHidden class. Adds a insetsHidden property to StyleModalStore to control whether this is checked or not.
- Adds logic in applyPanZoom to temporarily change the bounds of the SVG so it centers properly if insets are hidden. Add insetsHidden class to the svg and reload to test this. The PR to fix #46 will have this run on recenter.

Closes #207 and #199


apps/yapms/src/lib/components/modals/stylemodal/StyleModal.svelte
```diff
@@ -25,6 +32,16 @@
...
+			<h3 class="font-light text-lg">Other</h3>
+			<div class="flex flex-row gap-x-2 items-center">
+				<input
+					type="checkbox"
+					checked={$StyleModalStore.insetsHidden}
+					on:change={toggleInsets}
+					class="checkbox"
+				/>
+				<h3 class="font-light text-lg pb">Hide Insets</h3>
+			</div>
```
Fish1: Would using the daisyui checkbox example work here?
https://daisyui.com/components/checkbox/#with-label-and-form-control

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
```diff
@@ -48,4 +57,21 @@ function adjustStroke(
 	}
 }
 
+function boundsNoInset(svg: SVGElement) {
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: Maybe calling the functions "applyBoundsWithNoInset" or something would make it a bit clearer what the function is doing when being called in the code.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
```diff
@@ -1,16 +1,25 @@
 import panzoom, { type PanZoom } from 'panzoom';
 import z from 'zod';
 
+let originalBounds: DOMRect;
+let newBounds: DOMRect;
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: Could these be removed if the bounds were just computed inside the functions? The original bound can always be computed by getting the bounds of the SVG.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
```diff
@@ -1,16 +1,25 @@
...
 		const panzoomInstance = panzoom(svg, {
 			minZoom: 0.5,
 			maxZoom: 100,
 			smoothScroll: false,
 			autocenter: true,
 			zoomDoubleClickSpeed: 1
 		});
+		if (svg.classList.contains('insetsHidden')) {
+			boundsInset(svg);
+		}
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: Why are the bounds being set then unset here?

apps/yapms/src/lib/stores/Modals.ts
```diff
@@ -59,7 +59,8 @@ const ModeModalStore = writable({
 });
 
 const StyleModalStore = writable({
-	open: false
+	open: false,
+	insetsHidden: true
```
Fish1: The insets being hidden isn't really a property of the style modal store. Like how we have a ChartTypeStore and a ChartPositionStore, maybe we can have a MapInsetVisibilityStore.

apps/yapms/src/lib/components/navbar/NavBar.svelte
```diff
@@ -54,8 +54,11 @@
 	}
 
 	function openStyle() {
+		const svg = document.getElementById('map-div')?.querySelector<SVGElement>('svg');
+		const insets = svg ? svg.classList.contains('insetsHidden') : false;
 		StyleModalStore.set({
 			...$StyleModalStore,
+			insetsHidden: insets,
```
Fish1: I don't think this should be loaded via a class. Maybe it should be similar to the candidates attribute we have, where we load in some JSON then set the current application state.

Maybe we can seperate the default inset visibility state in another PR.

apps/yapms/src/lib/components/modals/stylemodal/StyleModal.svelte
```diff
@@ -10,6 +10,13 @@
 			open: false
 		});
 	}
+
+	function toggleInsets() {
+		const svg = document.getElementById('map-div')?.querySelector<SVGElement>('svg');
+		if (svg) {
+			svg.classList.toggle('insetsHidden');
+		}
+	}
```
Fish1: The modals themselves shouldn't directly modify the DOM, will explain in the final review.

Fish1 (changes requested):
Looks good and seems to be working from my testing! Just a few code reorganizing things.

The state of the modal visibility is currently stored inside the DOM itself. This should be broken out into its own store. Currenly you set the DOM and then have the DOM update the store. Where it should be the opposite, update the store, which then updates the DOM.

You can look at the Regions store for an example of how to use Store subscriptions. Basically the subscription will execute when the store changes, this is where you can then modify the DOM.

This should be stored in a new store that isn't specific to the style modal.
image

Also, kinda attached to this idea, we could move the default inset visibility state to a JSON attribute similar to how the candidates are loaded. But I think that can be a different PR.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: We need bounds to be set to no insets for panzoom to zoom in on regions specifically, but then we need them to be set back to include insets because if the user turns on insets, they will be cut off if bounds are not restored to the original. 

image

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/ApplyPanZoom.ts]: We could move newBounds in. 

For originalBounds, the bounds of the SVG are dictated by the height and width attributes, so, when those change, the bounds returned change as well:

image
Bounds stored in originalBounds on top, bounds returned from getBoundingClientRect() on bottom.

Flizzyclone:
Alright this PR has been reworked to update the DOM through a store and not the other way around. Please re-review.
- There is a new store, MapInsetsStore, that is created on map load. This is checked instead of checking whether or not the map SVG has insetsHidden class.
- The insetsHidden class is conditionally applied to #map-div based on this store.
- The Navbar updates this store, not the DOM itself.
- Checkbox has been updated to use form-control as well.
```
const insetsHidden = z.enum(['true', 'false']).safeParse(insetsHiddenAttribute);
if (insetsHidden.success) {
	MapInsetsStore.set({
		hidden: insetsHidden.data === 'true'
	});
```

The above code looks a bit silly- why check for a true false enum where you could check if `insets-hidden === "true"` with zod? Well, you'll never get an error message telling you why your insets aren't hiding that way since `insets-hidden === "true"` will always be some sort of boolean. This way, if you enter "tru" for example, you will get the correct error in the console.

Flizzyclone:
This PR now deals purely with adding insets, no logic for centering is involved.

Updated Summary of Code Changes:
- Adds MapInsetsStore to store whether or not insets are hidden
- Adds the insets-hidden class to map-div based on MapInsetsStore
- Adds a checkbox to toggle MapInsetsStore.hidden to the style modal
- Adds CSS rules to maintain background contrast on inset text & lines
- Adds CSS rules to hide inset regions when map-div has the insetsHidden class

Fish1 (approved):
LGTM, I just changed the CSS variables to all use dashes.

PR #210, opened 2023-07-04
Check saved maps for strings and valid regex.

Fish1:
With zod strings we can ensure that only approved characters go into the URL, and we don't need to manage an enum list.

### DIFF
--- a/apps/yapms/src/lib/types/SavedMap.ts
+++ b/apps/yapms/src/lib/types/SavedMap.ts
@@ -4,9 +4,9 @@ import { SavedRegionSchema } from './Region';
 
 export const SavedMapSchema = z.object({
 	map: z.object({
-		country: z.string().regex(/^[a-z]+$/),
-		type: z.string().regex(/^[a-z]+$/),
-		year: z.string().regex(/^\d{4}$/)
+		country: z.string().regex(/^\w+$/),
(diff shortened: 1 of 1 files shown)


apps/yapms/src/lib/types/SavedMap.ts
```diff
@@ -2,17 +2,11 @@ import { z } from 'zod';
...
-const years = ['2022'] as const;
-
 export const SavedMapSchema = z.object({
 	map: z.object({
-		country: z.enum(countries),
-		type: z.enum(types),
-		year: z.enum(years)
+		country: z.string().regex(/^[a-z]+$/),
+		type: z.string().regex(/^[a-z]+$/),
+		year: z.string().regex(/^\d{4}$/)
```
Flizzyclone: year needs to be more open ended. Routes like usa/house/blank or usa/senate/class1 exist that would not be covered under this regex.

Flizzyclone (changes requested):

apps/yapms/src/lib/types/SavedMap.ts
Fish1: I'll make an issue to go through and fix the naming for that then.

Flizzyclone (approved):
LGTM

PR #214, opened 2023-07-07
Set body width using percent

### LINKED ISSUE #102: Home page cuts off on mobile
We have identified this issue before, and implemented a fix that works in DevTools but not on my Pixel itself.

I've had this problem with my personal site before- I'll look into how I fixed it then and will try and get a fix working.
image

Flizzyclone:
This PR fixes an issue where the bottom of a page would be cut off on (certain non-gesture?) mobile devices. The best explanation for this that I have is that 100vh doesn't take into account either the top or bottom bar. I set body height and width to 100% instead of vw and vh and this problem goes away with seemingly no adverse affects on desktop or mobile.

Behavior before: 
Screenshot_20230707-002309
Screenshot_20230707-002323

Behavior after:

Summary of Code Changes:
- Switches the width and height css rules for html and body to use percent
- Adds some padding to the bottom of the div containing home page content for visual appeal

Closes #102

### DIFF
--- a/apps/yapms/src/lib/styles/global.css
+++ b/apps/yapms/src/lib/styles/global.css
@@ -4,8 +4,8 @@
 
 html,
 body {
-	width: 100%;
-	height: 100%;
+	width: 100dvw;
+	height: 100dvh;
 	overflow: hidden;
 }
 
--- a/apps/yapms/src/lib/styles/global.css
+++ b/apps/yapms/src/lib/styles/global.css
@@ -4,8 +4,8 @@
 
 html,
 body {
-	width: 100vw;
-	height: 100vh;
+	width: 100%;
+	height: 100%;
 	overflow: hidden;
 }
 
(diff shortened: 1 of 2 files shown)


Flizzyclone:
I've confirmed that vh doesn't account for the address bar: https://developer.chrome.com/blog/url-bar-resizing/

apps/yapms/src/lib/styles/global.css
```diff
@@ -4,8 +4,8 @@
 
 html,
 body {
-	width: 100vw;
-	height: 100vh;
+	width: 100%;
+	height: 100%;
```
Fish1: Should use dvw and dvh here, they dynamically adjust based on if the toolbars are shown or not.

https://web.dev/viewport-units/

Fish1 (changes requested):

Flizzyclone:
Changed to dvw/dvh w/o issue on mobile or desktop. Please re-review

Fish1 (approved):
LGTM

PR #217, opened 2023-07-07
Replace candidate-id, candidate-margin with candidates prop

### LINKED ISSUE #200: Use of Delegate edit feature not being preserved on export
I made a Multi-Member-District USA map (On the vein of the one made by fairvote [here](https://fairvote.org/sample-fair-representation-act-maps/)) and after using delegate edit to change the house rep counts for all the districts, I made a full house map:

Captura de tela 2023-06-27 191959

But after using the export button, all the multi-member districts I made returned to being single-member seats, and also got un-filled:

Captura de tela 2023-06-27 192039

The only thing unusual is that I imported a bunch of GeoJSONs from DRA and let YAPMS merge all those imports for me, but I don't think that should even be particularly important to the exporting process.

Flizzyclone:
This PR replaces the candidate-id and candidate-margin attributes previously used in SVG to set control of regions with the candidates (no relation to candidates attribute on the SVG, rename one?) attribute, which follows the following format:
```
export const RegionCandidatesSchema = z
	.object({
		candidate: z.string(), //Candidate ID
		count: z.number().nonnegative(),
		margin: z.number().nonnegative()
	})
	.array();
```

The candidate ID is then replaced with the corresponding Candidate object before the candidates property in Region object is set to the array passed in the attribute.

If anything is wrong with the value passed to the attribute, the region will default to the tossup candidate.

Summary of Code Changes:
- Adds RegionCandidatesSchema for zod validation of this attribute and a dervied type.
- Adds logic to parse the candidates attribute (getCandidatesForRegion)
- Updates export logic to adapt to the new attribute, also adds in support for disabled/locked
- Updates all maps to use the new attribute.
  - For most maps, this was a simple replace from `candidate-id="0"` to `candidates='[{"candidate":"0","count":1,"margin":0}]'`


apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -65,6 +66,36 @@ function findCandidate(id: string | null) {
 	}
 }
 
+function getCandidatesForRegion(candidateStr: string | null, value: number) {
+	if (candidateStr === null) {
+		return [{ candidate: get(TossupCandidateStore), count: value, margin: 0 }];
+	}
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: I think the function can simply not be called if there is no candidate string.

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -65,6 +66,36 @@ function findCandidate(id: string | null) {
...
+	try {
+		const parsedArr: RegionCandidates = JSON.parse(candidateStr);
+		RegionCandidatesSchema.parse(parsedArr);
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: You should use the schema parse to get the type here. Technically, even though it's not being used the parsedArr isn't of typeRegionCandidates when it is created.

image

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
```diff
@@ -65,6 +66,36 @@ function findCandidate(id: string | null) {
...
+				candidate: findCandidate(elem.candidate)
+			});
+		});
+		if (totCount !== value) {
+			candidateArr.push({
+				candidate: get(TossupCandidateStore),
+				count: value - totCount,
+				margin: 0
+			});
+		}
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: This will cause the tossup count to go negative. Would it be better to just return the tossup candidate and console log an error?

image

Fish1 (changes requested):
In addition to the comments, could you just have the variable names fully typed out without shorthand? Like parsedArr as parsedArray.

nitpick: Instead of parsedArr and Elem, they could be regionCandidates and candidate. It might be easier for future readers to understand what a variable is for.

image

apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/[map]/[year]/initialize/LoadRegions.ts]: Technically it should never hit this case- I put this here to avoid a bit of a messy conditional, but this code works as well and is nicer, will revise:

```
const candidateString = childHTML.getAttribute('candidates');
candidates: candidateString !== null //God bless our linting overlords.
	? getCandidatesForRegion(candidateString, value)
	: [{ candidate: tossupCandidate, count: value, margin: 0 }],
```

Flizzyclone:
Comments addressed including nit, please re-review.
Tossup count will no longer go negative & will throw an error and return tossup if totCount > value. If totCount < value remaining value will be filled with tossup candidate still.

Fish1 (approved):
LGTM

PR #221, opened 2023-07-08
Remove import warning

### LINKED ISSUE #193: Remove warning on import
As long as DomPurify is implemented correctly, I think we can remove this warning.

image

Removing possible script tags, should make it safe.

Flizzyclone:
This PR removes the warning from import modal as we now sanitize w/ DOMPurify

Closes #193


Fish1 (approved):
LGTM

PR #225, opened 2023-07-11
Use attributes for text selectors

Fish1:
Code: Converted any classes to data- attributes, for query selection. The for attributes was changed to data-for-region to be more semantically correct.

class .region-texts changes to attribute data-region-texts
class .bottom changes to attribute data-value-text
attribute for changes to attribute data-for-region
class .top is removed


Flizzyclone (approved):
I adapted non-usa maps. LGTM now

PR #227, opened 2023-07-12
Reformat usa presidential map

Fish1:
Code: The SVG data was all on a single line, making it difficult to modify by hand. This should simply reformat the map to be easier to edit with a text editor.


Flizzyclone (approved):
LGTM

PR #228, opened 2023-07-13
Remove prefixes & "sticky" modal

### LINKED ISSUE #224: Remove the name of the country from the modals
On the more maps modals, we don't need the name of the country for every button. The modal itself implies that the map is of the the country it is for.

image

Flizzyclone:
This PR adds a "prefix" property to cards in MapSelection.json that is used to filter out the specified text in titles in MoreMapsModal. It also adds an optional "sticky" property to ModalBase that makes the title and action bars stick in place as you scroll.

Summary of Code Changes:
- Adds "prefix" to MapSelection.json
  - Adds a prefix property to MoreMapsModalStore
  - Adds logic for the prefix to be replaced in MoreMapsModal
  - Sets prefixes for existing cards
- Adds sticky property to ModalBase
  - Adds conditional classes to ModalBase to make title/action bars sticky
  - Makes MoreMapsModal 'sticky'

### DIFF
--- a/apps/yapms/src/lib/assets/homedata/MapSelection.json
+++ b/apps/yapms/src/lib/assets/homedata/MapSelection.json
@@ -8,7 +8,6 @@
 				"alt": "U.S Capitol",
 				"doubleCols": true,
 				"attribution": "Photo by Andy Feliciotti",
-				"prefix": "USA",
 				"links": [
 					{
 						"label": "President",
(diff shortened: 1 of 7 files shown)


Fish1 (changes requested):
The stick buttons on the modal seem good to me!

For the titles, I think we should avoid doing the prefix manipulation. It probably wont be a huge issue. But it's maintaining an inter-dependency between two parts of the program that don't need to be connected (the title on the app page, and the label on the button link).

I think if we split the button label into it's own thing it will give us more control over the label and the page titles. Maybe the label for the button can be added to the routes array alongside the link.

(if you do this I think the routes array can be renamed to buttons or something)

image

Flizzyclone:
Ahh this is a good idea. Will change

Flizzyclone:
Alright I reworked MoreMapsModal to use the buttons structure as you described. I reused the HomeLinkData type for the buttons property.

Summary of Code Changes:
- Changes MapSelection.json to use buttons prop (HomeLinkData[]) for modals instead of a string[] of routes
- Instead of taking an array of routes and searching Titles.json, MoreMapsModal will now take an array of HomeLinkData and display title and route based on that
- Passes buttons prop to MoreMapsModalStore

Fish1 (approved):
LGTM

PR #230, opened 2023-07-13
Tell Svelte which routes to render

Flizzyclone:
This PR fixes a bug where going to a route not located directly on the home page would return a 404. For instance https://yapms2.com/app/de/folketing/blank will return a 404.

This is caused by Svelte not prerendering this routes. If we remove prerendering for [country]/[map]/[year], it is fixed, but we lose speed as clients have to build the page themselves. This PR tells Svelte which routes to prerender based on Titles.json, which will include any map a user would ever hit.

To test this, do `npm run build` in apps/yapms followed by `npm run preview` then navigate to a route that doesn't appear on the home page.

Summary of Code Changes:
- Adds an [entries](https://kit.svelte.dev/docs/page-options#entries) function for [country]/[map]/[year] to tell Svelte what to prerender.


Fish1 (approved):
LGTM

PR #231, opened 2023-07-13
Candidate ID fix

Flizzyclone:
This PR fixes an issue on the Dutch House map where BVNL had the same candidate ID as BIJ1


Fish1 (approved):
LGTM

PR #235, opened 2023-07-15
Fix map oversensitivity on mobile

### LINKED ISSUE #229: Coloring
Whenever I try to move the screen around on mobile, it auto fills if I touch it (even when scrolling). In yapms original, it does not do that, so maybe make it a bit less sensitive or at least an option to choose.

Flizzyclone:
This PR changes our click event to onclick and takes a panzoom setting from yapms1 (seemingly inspired by [this documentation](                })) to make sure that click actually works on mobile. This fixes #229 where the map would fill when you went to pan the map.


Fish1 (approved):
Works for me!

CenterRightPatriot:
Thanks

PR #241, opened 2023-07-16
Prevent closing of MoreMapsModal when opening in new tab

Flizzyclone:
This PR prevents MoreMapsModal from closing when a link is opened in a new tab using the control key. You can open multiple maps quickly this way.


Fish1 (approved):
LGTM

PR #242, opened 2023-07-16
Move chart logic to common layout

Fish1:
Code: Pulls the chart area out into its own component, and puts it into the common layout.


Flizzyclone (commented):
LGTM

Flizzyclone (approved):

PR #243, opened 2023-07-17
USA Historical Presidential Maps

Flizzyclone:
This PR adds US presidential election maps from 1788 all the way to 2020. I disabled states that weren't in the union yet/reconstructed instead of removing them from the map.

All maps are derived from 2022 presidential w/ some changes to the ME and NE districts sourced from Wikipedia maps for that decade. All results (and most colors) were obtained from Wikipedia as per usual. The picture is from [Pexels](https://www.pexels.com/photo/the-white-house-in-autumn-11563771/). CC0 as always.


Fish1 (approved):
LGTM

PR #247, opened 2023-07-20
Auto borders for all maps

Flizzyclone:
This PR changes all maps to using auto borders. Some of these required me to make source changes.

Denmark Folketing source changed from wikipedia to UMich's [CLEA](https://electiondataarchive.org/data-and-documentation/georeferenced-electoral-districts-datasets/)

All the changes in this PR are changes to paths to remove the offset borders they used before and adding auto border arguments.


Fish1 (approved):
LGTM

PR #249, opened 2023-07-22
reapplyPanZoom on chart changes

### LINKED ISSUE #46: Odd map behavior when changing chart position.

Flizzyclone:
This PR adds logic to reapply panzoom and reapplies panzoom when chart options are changed.

Summary of Code Changes
- Moves svg & panzoomInstance variables from function scope to file scope. Changes applyPanZoom to accommodate
- Creates new function storeSVGForPan to set the SVG that panzoom should be applied to.
- Creates reapplyPanZoom to dispose of current panZoom instance and create a new one
  - applyPanZoom is wrapped in a setTimeout in order to make sure the DOM updates before it runs.
- Calls reapplyPanZoom when chart options are changed

Fixes #46


Fish1 (approved):
LGTM

PR #251, opened 2023-07-23
Consolidate custom attributes

Fish1:
Code: Moved all classes to custom attributes. Also removed the data- prefix from ones that had it.


Flizzyclone (approved):
Did a two fixes related to the import feature. LGTM.

PR #255, opened 2023-07-25
Rework Map URLs

### LINKED ISSUE #215: Canadian commons route needs a year
There should be a year for the Canadian house of commons map, instead of just blank.

image

### LINKED ISSUE #245: All maps need a date, or unique identifier
The maps are saved with the following...
image

So if the year part just says current or blank, then the map changes, all the saved maps for people will be corrupted.

Flizzyclone:
This PR changes map URLs from [country]/[map]/[year] to either [country]/[map]/[year]/[varient] or [country]/[map] for maps without a year. I will be referring to these two types of URL as 4-param and 2-param respectively.

The first &, second & third commits are likely to be the most useful when reviewing. Other commits have to do with renaming files and data updates.

For naming of maps I tried to use 4-param as much as possible but for maps without a specific year I fell back to 2-param. For maps like the Canadian House of Commons where boundaries have been the same since 2011 I used the most recent election date for the year parameter.

Summary of Code Changes 
- Moves [country]/[map]/[year] +page.svelte to [country] +layout.svelte so that it is used for all subdirectories of [country]
  - Adds a blank +page.svelte into [country] subdirectories so we don't hit 404 when we access them.
- Adds variant param to redirects, saved map schema, & saved map objects for 4-param URLs
- Make variant param & year param optional in saved maps to support 2-param URLs
- Adds support for variant param & 2 param URLs to prerendering code
  - Prerendering code was broken out into two +page.server.ts files based on number of parameters.
- Gives most maps new URLs and standardizes names in Titles.json

Closes #245 and #215


Flizzyclone:
We could probably makes the conditionals for 2-param and 4-param less explicit. So instead of two different sets of logic it can be one set and the other two params simply don't get added if they don't exist.

Fish1 (approved):
LGTM, feel free to refactor in a new PR if you'd like.

PR #258, opened 2023-07-26
Update Major Versions

Fish1:
Updated svelte, prettier and some eslint rules to their next major versions.

I had to remove the placeholders from the searchbar on the homepage because it doesn't actually have a "placeholders" property to set the placeholder. I am not sure how it got past CI before.


Flizzyclone (approved):
LGTM

PR #270, opened 2023-07-30
Fix duplicate county IDs

Flizzyclone:
This PR fixes an issue on the US counties map where some regions had duplicate region IDs so they could not be filled.


Fish1 (approved):
LGTM

PR #275, opened 2023-08-03
Moved Edit Candidate Button

Fish1:
Feature: The edit button is moved to this new candidate modal. With the modal you can click a candidate to edit, or click the add button to add.

In the future I think we can make this modal have the ability to reorganize the candidate via drag and drop or something.

image

Code:Added a base candidate modal in which you can navigate to the other modals.


Flizzyclone:
Two nits with this, but that is it.

1. I am not a fan of the glass on candidate buttons, it messes with contrast & as inconsistent with the rest of the app.
2. I think the "Add" button should be changed to "Add Candidate". We have the space for it and it just makes it a bit more descriptive.

Flizzyclone (approved):

Fish1:
Yeah, I was trying to figure out why the corners of the buttons looked messed up. I just needed to add btn-ghost to it. The button color properties will clean up the borders.

I disagree with adding the word "candidate" to the add button. Because I think within the context of a list of candidates, "add" implies adding to the list of candidates.

Like how the add button under the colors here, implies it's adding to the list of colors. And update implies that it updates the candidate. 

And yes I notice that we have the "remove candidate" button. I do think that should be moved to a better position where it has a more contextual meaning so it doesn't need to say "remove candidate". Either next to the update button, or maybe somewhere on the list of candidates.
image

PR #276, opened 2023-08-03
Take-All & Territories

Flizzyclone:
This PR adds 2 maps:
- USA 2024 presidential take-all 
- 2024 presidential with territories

Both maps are derived from USA 2022 presidential.


Fish1 (approved):
LGTM

PR #282, opened 2023-08-06
Fix to USA Senate disabling

Flizzyclone:
This PR fixes an issue with U.S Senate maps that have circles to indicate second seats. The primary state did not have the secondary circle cut out of it, resulting in it appearing that nothing had changed when you disabled the second seat. This PR updates the primary states to have the circle cut out.

image
Behavior before

image
Behavior after


Fish1 (approved):
LGTM

PR #285, opened 2023-08-06
Proper Pie Sizing

Flizzyclone:
This PR fixes an issue where the pie chart would be too small if type was still 'battle' but pie was rendered due to over 2 candidates.


Fish1 (approved):
LGTM

PR #287, opened 2023-08-07
Better modal interactivity

Flizzyclone:
This PR changes modals to take advantage of the accessibility improvements of the dialog component (better focusing for screen readers, can close by clicking outside, can close using escape key)

Summary of Code Changes:
- Removes open property & checkbox
- binds dialog element to dialog variable, calls showModal and close on it when the open variable in store changes
- Adds an event listener to call close function when modal is closed by pressing escape
- Adds a modal-backdrop form with a button inside to handle clicking outside.

### DIFF
--- a/apps/yapms/src/lib/components/modals/ModalBase.svelte
+++ b/apps/yapms/src/lib/components/modals/ModalBase.svelte
@@ -4,7 +4,7 @@
 
 	export let title: string;
 
-	export let store: Writable<{ open: boolean }> | undefined;
+	export let store: Writable<{ open: boolean }>;
 	export let onClose: (() => void) | undefined = undefined;
 
 	export function close() {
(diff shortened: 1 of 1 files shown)


apps/yapms/src/lib/components/modals/ModalBase.svelte
```diff
@@ -14,18 +13,19 @@
 		} else if ($store !== undefined) {
 			$store.open = false;
 		}
-		console.log(open);
 	}
 
 	let content: HTMLDivElement | undefined;
 	let offsetHeight: number;
+	let dialog: HTMLDialogElement;
```
Fish1: This just needs to be of type undefined, as it's undefined at initialization. The offsetHeight above it probably also needs to be of type undefined.

apps/yapms/src/lib/components/modals/ModalBase.svelte
```diff
@@ -14,18 +13,19 @@
...
 
 	$: isOverflow = offsetHeight < (content?.scrollHeight ?? 0);
-</script>
 
-<input type="checkbox" class="modal-toggle" checked={$store?.open ?? false} />
+	$: if ($store?.open) dialog?.showModal();
+	$: if (!$store?.open) dialog?.close();
```
Fish1: The store types undefined type can be removed, as it's required to be in all modals to function.

Fish1 (changes requested):

Flizzyclone:
Types fixed. Please re-review

Fish1 (approved):
LGTM

PR #288, opened 2023-08-07
fix NZL "Area Outside"

Flizzyclone:
This PR removes the "Area Outside" region on the New Zealand regional councils map. Turns out that although it is included in the shapefile, it is not an actual Regional Council. Thanks to BlockyBeach for pointing this out.


Fish1 (approved):
LGTM

PR #292, opened 2023-08-10
Added auto stroke to the /view route

### LINKED ISSUE #257: Borders not being applied in the /view route
There doesn't appear to be borders for the screenshots.

image

Fish1:
Feature: The auto stroke wasn't being applied to the /view route. I refactored and created an applyAutoStroke function that can be used in the /view route.


Flizzyclone (approved):
LGTM

PR #294, opened 2023-08-10
Remove data prefixes

### LINKED ISSUE #262: Attributes using data prefix in maps
Some maps are still using the data prefix for region texts, throwing their behavior off.

Flizzyclone:
This PR removes data- on attributes that no longer use the data prefix. This problem existed on Danish folketing maps.

Closes #262


Fish1 (approved):
LGTM

PR #295, opened 2023-08-11
Use forms for text input

### LINKED ISSUE #284: Adapt modals to use forms
Modals where the user is entering data should be using form so that you can submit by pressing enter

This was brought up by [Luna](https://discord.com/channels/1038861619540795512/1038875719666765954/1137186577597091901)

Flizzyclone:
This PR changes text inputs in modals to use <form> elements so that you can press enter to submit them when focused. EditRegionModal is also modified to have a label & error text for it's input as well as different handling for invalid values.

- Wraps all modal text inputs in a form with on:confirm={submit}
- Adds error text for EditRegionModal when a negative number is entered.
- Disables EditRegionModal's confirm button on invalid input. When pressing enter on an invalid input, you will get a popup and it will not submit: 
image

Closes #284

### DIFF
--- a/apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
+++ b/apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
@@ -34,9 +34,7 @@
 			<h3 class="font-light text-lg">Region Value</h3>
 			<form on:submit={confirm}>
 				<input
-					id="editRegionInput"
 					type="number"
-					placeholder="Value"
 					class="input input-bordered w-full"
 					on:keypress={preventNonNumericalInput}
 					bind:value={newValue}
--- a/apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
+++ b/apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
@@ -8,7 +8,9 @@
 	$: value = open ? $EditRegionModalStore.region?.permaVal : undefined;
 	$: newValue = value ?? 0;
 
-	$: negativeNumError = newValue < 0;
+	function preventNonNumericalInput(e: KeyboardEvent) {
+		if (!e.key.match(/^[0-9]+$/)) e.preventDefault();
+	}
 
 	function confirm() {
 		if (newValue === null) {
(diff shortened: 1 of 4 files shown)


Fish1 (commented):
I don't think I like the enter button being on the candidate modals. Because after the user enters the name of the candidate, there are still other actions that may need to be completed. And there is no way for us to know if the user is actually done filling the form or not.

I think it does make sense in the context of the edit region modal though, because we know the user is actually done filling out the form.

apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
```diff
@@ -8,9 +8,8 @@
 	$: value = open ? $EditRegionModalStore.region?.permaVal : undefined;
 	$: newValue = value ?? 0;
 
-	$: if (newValue < 0 || newValue === null) {
-		newValue = 0;
-	}
```
Fish1: Are you against preventing the user from entering invalid data here?

Flizzyclone: I am against modifying the contents of newValue directly. Under that behavior, whenever the user deletes the current value to enter in a new value, a 0 will be put in that they can't remove.
yapmsZeroInsert

Preventing invalid data from being submitted with an explanation of why that data is invalid seems like the better solution to me.

Flizzyclone:
Removed the forms on candidate modals, please re-review

apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
Fish1: Oh didn't notice that.

What if we just allowed nulls, and treated it as 0 when the user submitted?

Flizzyclone: That sounds good. Will change

Flizzyclone:
Changed null to being handled on submit. Removed required on input element to allow for submitting null using the enter key.

apps/yapms/src/lib/components/modals/editregionmodal/EditRegionModal.svelte
Fish1: I was thinking to keep the behavior of preventing the user from entering negative numbers. Are you against that?

Fish1: Also, I am able to type random stuff into the box and it allows it.

image

Flizzyclone: I will fix random stuff, imo having a descriptive error message telling the user what is wrong when they enter in a negative number is good. We could change it to warning text that says "negatives are not supported" and then when it is submitted just take the absolute value.

Fish1: Alright then, the current message you have is fine.

Flizzyclone:
Entering letters on Firefox is fixed

Flizzyclone:
Removed the negative check entirely as the prevent function will prevent typing in "-" in the first place. Also removed min="0" on the input for this reason

Fish1 (approved):
LGTM 

Just removed the id and placeholder since the headers already there.

PR #298, opened 2023-08-13
New Zealand clip fix

Flizzyclone:
New Zealand house maps had part of the Northland region clipped off. This PR changes the d prop for Northland to fix this region.


Fish1 (approved):
LGTM

PR #299, opened 2023-08-13
Lock Map Movement

### LINKED ISSUE #220: Lock map in place
In yapms1 you can lock the map in place. We should probably implement that here.

Fish1:
Feature: added a lock map feature in the options menu. Also renamed chart options to just options.

I think I plan on changing the buttons to selects instead, and it might look at bit better.

Code: Subscribe to the LockMapStore, when it changes, then pause panzoom.

closes #220


apps/yapms/src/lib/components/modals/optionsmodal/OptionsModal.svelte
```diff
@@ -57,6 +58,12 @@
 					</button>
 				{/each}
 			</div>
+			<div class="form-control">
+				<label class="label cursor-pointer">
```
Flizzyclone, line 62: We will need to decide on consistent styling for these types of checkbox inputs. Looks different in style modal. This can happen once we unify stuff from style modal into options.

Flizzyclone (approved):
LGTM

PR #302, opened 2023-08-14
Edit region paste check

Flizzyclone:
This PR adds in a function to check that content being pasted into the edit region input matches correct format (0-9). Also adds min="0" back to the input so that the side arrows cannot make input value a negative number.


Fish1 (approved):
LGTM

PR #303, opened 2023-08-14
Remove style in favor of options & theme

### LINKED ISSUE #293: "None" chart logo
Right now there is no way to remove the logo once you've added it.

Flizzyclone:
This PR removes the style modal, moving themes to thememodal (as used in the home page) and other style options such as logo to the options modal. Options modal also gets a redesign here in order to fit all the options in.

image

Summary of Code Changes
- Adds logic in OptionsModal to change logos
- Creates selects & checkboxes for options besides logos. Binds value to stores for easy updating.
- Deletes StyleModal
- Adds ThemeModal to layout.svelte, adds a button for Theme to the navbar.


apps/yapms/src/lib/components/modals/optionsmodal/OptionsModal.svelte
```diff
@@ -1,67 +1,83 @@
...
 					>
-						<span>
-							{type}
-						</span>
-					</button>
-				{/each}
+						{#each logos as logo}
+							<option>{logo.name}</option>
+						{/each}
+					</select>
```
Fish1: Is there a reason for not binding the LogoStore similar to the other selects here?

Fish1 (commented):
Looks great. 

Just wondering why the LogoStore wasn't binded like everything else.

apps/yapms/src/lib/components/modals/optionsmodal/OptionsModal.svelte
Flizzyclone: Yeah. If we bind Logostore to value, Logostore will become the logo name and not the src like it needs to be. We could change up the store to have the logo object stored instead of just src, but I didn't do that to keep the PR small.

Flizzyclone: Thanks to the reactive statement it should update the select when Logostore changes just as if it was binded

Fish1: I think this should solve the issue, you can set the value of the option to the source, while keeping the display value as the name.

```html
<select class="select select-bordered w-full" bind:value={$LogoStore}>
  {#each logos as logo}
    <option value={logo.src}>{logo.name}</option>
  {/each}
</select>
```

Flizzyclone: Oh good call! Changed logos to work this way.

Fish1 (approved):

PR #308, opened 2023-08-17
Fix Canadian "Delta" region

Flizzyclone:
The non-inset path for the Canadian "Delta" region on Commons maps would disappear when insets were toggled off. This PR fixes that.


Fish1 (approved):
LGTM

PR #309, opened 2023-08-17
Two-Way battle chart ordering fix

Flizzyclone:
This PR fixes an issue where the battle chart margins wouldn't be aligned to the middle on a two-way chart.

After changes:
image

Summary of Code Changes:
- Reverses the way in which CandidateCountsMargins and CandidatesStore elements are accessed in the battle chart


Flizzyclone:
#310 includes refactor, making this redundant.

PR #310, opened 2023-08-18
Reverse sort of 2nd candidate on battle chart

Fish1:
Reversed the order of the counts on the 2nd candidate of the chart.

Also, refactored the code to be more understandable.


Flizzyclone (approved):
LGTM

PR #312, opened 2023-08-18
Remove chart change when candidates are added

### LINKED ISSUE #311: Don't change chart type when more candidates are added.
Since the battle chart can function with more than a single candidate, don't switch.

Fish1:
Removed the logic causing the chart to change when more candidates are added.


Flizzyclone (approved):
LGTM

PR #316, opened 2023-08-21
Fix zoom on ios

Fish1:
This change is to prevent the pinch action from zooming in the entire web browser. Now if the user pinches on the map, only the map will zoom in.

### DIFF
--- a/apps/yapms/src/lib/utils/applyPanZoom.ts
+++ b/apps/yapms/src/lib/utils/applyPanZoom.ts
@@ -18,20 +18,34 @@ function lockMap(lock: boolean) {
 	}
 }
 
-function applyPanZoom(svg: SVGElement) {
+function applyPanZoom(svg: SVGSVGElement) {
 	if (panZoomSettings !== undefined) {
 		panZoomSettings.panzoom.dispose();
 	}
+	const bb = svg.getBBox();
+	svg.setAttribute('viewBox', '0 0 ' + (bb.x + bb.width + bb.x) + ' ' + (bb.y + bb.height + bb.y));
+	let handler: NodeJS.Timeout | undefined;
+
 	const panzoomInstance = panzoom(svg, {
-		minZoom: 0.5,
-		maxZoom: 100,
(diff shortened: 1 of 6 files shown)


apps/yapms/src/lib/utils/applyPanZoom.ts
```diff
@@ -18,17 +18,34 @@ function lockMap(lock: boolean) {
...
-		maxZoom: 100,
+		transformOrigin: { x: 0.5, y: 0.5 },
 		autocenter: true,
 		zoomDoubleClickSpeed: 1,
 		smoothScroll: false,
-		onTouch: function () {
+		initialX: (svg.parentElement?.offsetWidth ?? 0) / 2,
+		initialY: (svg.parentElement?.offsetHeight ?? 0) / 2,
+		initialZoom: 0.85,
+		zoomSpeed: 0.05,
```
Flizzyclone, line 37: I don't know if this is required for iOS or anything, but this is really slow on desktop. Could we up this again?

apps/yapms/src/lib/utils/applyPanZoom.ts
```diff
@@ -18,17 +18,34 @@ function lockMap(lock: boolean) {
...
+		onTouch: (event) => {
+			if (event.touches.length === 2) {
+				event.preventDefault();
+			}
+			console.log(panzoomInstance.getTransform().scale);
+			clearTimeout(handler);
+			handler = setTimeout(() => {
+				if (svg.parentElement) {
+					//svg.parentElement.style.transform = `scale(${1 / panzoomInstance.getTransform().scale})`;
```
Flizzyclone, line 46: Does this timeout need to be here if this is commented?

Flizzyclone (changes requested):

apps/yapms/src/lib/utils/applyPanZoom.ts
Fish1: This is still a WIP, I am just trying different things.

PR #317, opened 2023-08-24
Removed disabled share modal

Fish1:
Feature: Moved the export SVG button to the import map. We can remove the Disabled Share Modal with this.


Flizzyclone (changes requested):
So the only thing I would ask for here is that a tooltip be added around the Generate Link button to explain why it is disabled.

```
<div class="tooltip-right" class:tooltip={$page.url.pathname === '/app/imported'} data-tip="Link Generation is disabled on imported maps.">
  <button
    class="btn btn-secondary gap-1 flex-nowrap"
    on:click={generateLink}
    disabled={fetchingLink ||
	    turnstileToken === null ||
	    $page.url.pathname === '/app/imported'}
    >
    <Link class="w-5 h-5" />
    <span>Generate Link</span>
  </button>
</div>
```

Fish1:
Just working on trying to allow the tooltip to overflow the modal. It seems to be getting cut off for some reason.

PR #318, opened 2023-08-25
Initial Documentation App

Fish1:
Added the starter astro app with the starlight theme. It's an official theme from astro, so I hope it gets enough support and we don't need to worry about themes ourselves so much.


Flizzyclone (approved):
LGTM

PR #321, opened 2023-08-27
Dynamic Links

### LINKED ISSUE #175: Make social media links dynamic
Social media links should be inserted at build time or queries from pocketbase, as they can change. Discord links often change.

Flizzyclone:
This PR allows us to dynamically define links for things like social media, other tools, etc.

Summary of Code Changes:
- Creates a new pocketbase collection for social links. 
- Creates new SocialLinkGrid component containing buttons for links from the social_links collection.
  - I tried dynamic imports here to avoid having to use `import *`. I could not get this to work and ultimately decided against this method as the dev server would crash if a non-existent module was passed (ex. you enter a name of an icon that doesn't exist)
- Updates social links to use SocialLinkGrid component

Closes #175


Fish1:
This isn't what I meant by dynamic. By dynamic I was thinking we could have a table on pocketbase where we could update links, and have users query them. This way we don't need to redeploy the application just because a discord invite expired or something.

Fish1:
If you'd just like to update the twitter icons, I think it would just be simpler to just keep it in the HTML.

Flizzyclone:
Ohh lol. I will rework this.

Fish1:
I think when I said build time, I was thinking about environment variables. That way the links weren't tracked under source control, and we could change it at that point.

But I think pocketbase would be more ideal.

Flizzyclone:
Alright so the idea behind importing all icons is so that if we want to add a new link, we can do that without having to re build.

image

image

Flizzyclone:
I have updated this to use PocketBase, please re-review.

Fish1:
I'll review this later tonight, but just skimming over it, I would probably see if the getLinks utility logic and just be put directly into the ExternalLinkGrid. So there isn't a need to pass promises around.

Fish1 (changes requested):
This could be simplified by simply removing the mount and getLinks function. The getFullList function already returns a Promise for an array of records. And you can catch errors by adding the :catch clause to the :await

I would also highly recommend removing the "Categories" from the links table and simply having a table dedicated to the social media links or something.

Because I would think different categories of links might have different data that we might want to store with them. I could see that quickly breaking normalization the moment we want to do something a bit different.


```svelte
<script lang="ts">
	import Fa from 'svelte-fa';
	import * as BrandIcons from '@fortawesome/free-brands-svg-icons';
	import * as SolidIcons from '@fortawesome/free-solid-svg-icons';
	import { PocketBaseStore } from '$lib/stores/PocketBase';

	export let category: string;

	let links = $PocketBaseStore.collection('links').getFullList({
		filter: `category = "${category}"`
	});

	function getIcon(name: string) {
		const brandIcon = BrandIcons[name as keyof typeof BrandIcons];
		const solidIcon = SolidIcons[name as keyof typeof SolidIcons];

		if (brandIcon !== undefined) {
			return brandIcon;
		} else if (solidIcon !== undefined) {
			return solidIcon;
		}
	}
</script>

<div class="flex flex-row flex-wrap gap-2 justify-center">
	{#await links}
		<span class="loading loading-ring"></span>
	{:then links}
		{#each links as link}
			<a class="btn btn-block btn-sm px-4 flex-nowrap flex-grow gap-2" href={link.URL}>
				<Fa icon={getIcon(link.faIcon)} />
				<span>{link.label}</span>
			</a>
		{/each}
	{:catch error}
		{error.status}
	{/await}
</div>
```

Flizzyclone:
Alright I have changed the links collection to be specifically for social-links, removed the category property, and updated the component to follow the code you posted & to only be for social links. Please re-review

Note on caching: I will make an issue for caching these once they are merged in. I need to read more on how browsers cache and how we could do this using browser functionality.

Fish1 (approved):
LGTM

PR #322, opened 2023-08-27
Added a display of the map sources

Fish1:
Feature: On the sidebar, sources will for the current map will be displayed.

The original-source attribute for all the maps need to have the strings double quotes, as single quotes are not valid in JSON.

image


Flizzyclone (approved):
LGTM

PR #325, opened 2023-09-03
Custom Colors

Fish1:
Feature: Added a custom colors tab. The user can add an unlimited number of custom colors and edit them.

Code: Added a few new modals to edit and add new colors. Refactored the color buttons out to its own component. Also added svelte local storage stores, so that the website can remember the custom colors.


Flizzyclone:
I suppose this would be a nitpick, but I would like to see some text here saying something like "You have no custom colors defined. To create a new slate, use the 'Add' button.

image

Also, we should make preset colors available when editing a candidate at some point.

Other than that nit, LGTM.

Flizzyclone (approved):

Fish1:
How about this?
image
image

PR #327, opened 2023-09-05
Fix Blurry Map on iOS

### LINKED ISSUE #261: Map Blurred / Border is not very clear
As you can see in the images below, the map I uploaded is extremely blurry, and I especially can’t color places like NYC or LA as they are too blurry. This only happens on yapms2 and not yapms, which could be due to the lines being smoother on yapms, and that the size doesn’t auto rechange, or it could be a graphics issue somewhere.

yapms2 imported map view:
26F1E42F-0221-4F87-91DF-959B8949A1FE
BA501D08-1736-4AAB-94B2-1AE4EC05C171

regular yapms
9874B861-B17B-4A31-A954-94937AA2C0D7

Flizzyclone:
This PR fixes the map on iOS being blurry when you zoom in. It does this by setting a z-index value for the navbar. lg:overflow-x-clip is also removed and I changed overflow-x-scroll to overflow-x-auto which will allow you to scroll still. Tested on Lambdatest & an Iphone SE (2020)

Closes #261 
#316 is also no longer needed.

### DIFF
--- a/apps/yapms/src/lib/components/navbar/NavBar.svelte
+++ b/apps/yapms/src/lib/components/navbar/NavBar.svelte
@@ -66,7 +66,7 @@
 </script>
 
 <div
-	class="navbar flex-row bg-base-200 gap-3 overflow-x-scroll overflow-y-clip lg:overflow-x-clip min-h-0"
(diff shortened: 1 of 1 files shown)


Fish1 (approved):
LGTM nice debugging!

PR #328, opened 2023-09-05
Pinch to zoom caused entire page to zoom

Fish1:
At least on my iPod, pinch to zoom caused the entire webpage to zoom in. This should prevent it.


Flizzyclone (approved):
LGTM

PR #346, opened 2023-09-08
Stripes for split states

Flizzyclone:
This PR changes equally split states to have their fill be a striped pattern rather than a single color. Works with any number of candidates.
image

I have to rely on some DOM manipulation of the SVG here because patterns can only be defined within SVG. It is annoying.

Summary of Code Changes
- Modifies Regions.ts to identify if multiple winners present, if so, set fill to pattern, if not, set fill to winner color.
- Adds type named "RegionCandidate" for the structure contained within the candidate property of a region. 
- Renames  "RegionCandidates" type and corresponding schema to "SavedRegionCandidates" 
perhaps we ought to refactor so those can use the same type. was out of scope of this PR.
- Adds blendHexForLuma function which takes in an array of hex codes and returns their blended color. For striped states, this is then passed to luma to get black or white text.
- Adds function removeAllPatterns to remove all patterns when regions is updated. This is done to avoid duplicate versions of a pattern for the same two candidates.
- Adds function makePattern to create a stripe pattern for the candidates passed to it, add that pattern to the DOM, and then return the pattern name for use in fill.
- Fixes an issue where not all candidate objects would be properly edited within regions when candidates were edited.

Also, if this PR never gets merged in I won't mind. I was going to let this feature die, but people keep asking about it, so I finished it. I don't particularly like the code here (I look forward to what you think of for improving it), but it gets the job done.

A potential enhancement in the future could be to toggle this on/off. I chose not to implement it here because it requires that I would force a re-render of regions, and I couldn't get that to work reliably under a store.subscribe.

### DIFF
--- a/apps/yapms/src/lib/types/Region.ts
+++ b/apps/yapms/src/lib/types/Region.ts
@@ -24,6 +24,12 @@ export const RegionSchema = z.object({
 	})
 });
 
+export const RegionCandidateSchema = z.object({
+	candidate: CandidateSchema,
+	count: z.number(),
+	margin: z.number()
+});
+
 export const SavedRegionSchema = RegionSchema.omit({
 	shortName: true,
 	longName: true,
@@ -39,7 +45,7 @@ export const SavedRegionSchema = RegionSchema.omit({
 		.array()
 });
 
-export const RegionCandidatesSchema = z
+export const SavedRegionCandidatesSchema = z
 	.object({
 		candidate: z.string(),
 		count: z.number().nonnegative(),
@@ -49,6 +55,8 @@ export const RegionCandidatesSchema = z
 
 type Region = z.infer<typeof RegionSchema>;
 
-export type RegionCandidates = z.infer<typeof RegionCandidatesSchema>;
+export type RegionCandidate = z.infer<typeof RegionCandidateSchema>;
+
+export type SavedRegionCandidates = z.infer<typeof SavedRegionCandidatesSchema>;
 
 export default Region;
--- a/apps/yapms/src/lib/utils/luma.ts
+++ b/apps/yapms/src/lib/utils/luma.ts
@@ -6,7 +6,7 @@ function hexToRGB(hex: string) {
 	return { r, g, b };
 }
 
-function blendHexForLuma(hexes: string[]) {
+function blendHexColors(hexes: string[]) {
 	let rCombined = 0,
 		gCombined = 0,
 		bCombined = 0;
@@ -22,7 +22,7 @@ function blendHexForLuma(hexes: string[]) {
 	rCombined = Math.floor(rCombined);
 	gCombined = Math.floor(gCombined);
 	bCombined = Math.floor(bCombined);
-	return rCombined.toString(16) + gCombined.toString(16) + bCombined.toString(16);
+	return '#' + rCombined.toString(16) + gCombined.toString(16) + bCombined.toString(16);
 }
 
 function calculateLumaRGB(r: number, g: number, b: number): number {
@@ -35,4 +35,4 @@ function calculateLumaHEX(hex: string): number {
 	return result;
 }
 
-export { blendHexForLuma, calculateLumaHEX, calculateLumaRGB };
+export { blendHexColors, calculateLumaHEX, calculateLumaRGB };
--- a/apps/yapms/src/lib/utils/luma.ts
+++ b/apps/yapms/src/lib/utils/luma.ts
@@ -6,6 +6,25 @@ function hexToRGB(hex: string) {
 	return { r, g, b };
 }
 
+function blendHexForLuma(hexes: string[]) {
+	let rCombined = 0,
+		gCombined = 0,
+		bCombined = 0;
+	for (const hex of hexes) {
+		const { r, g, b } = hexToRGB(hex);
+		rCombined += r;
+		gCombined += g;
+		bCombined += b;
+	}
+	rCombined /= hexes.length;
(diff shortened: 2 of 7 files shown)


apps/yapms/src/lib/types/Region.ts
```diff
@@ -39,7 +45,7 @@ export const SavedRegionSchema = RegionSchema.omit({
 		.array()
 });
 
-export const RegionCandidatesSchema = z
+export const SavedRegionCandidatesSchema = z
```
Fish1, line 48: Why was this renamed to "SavedRegionCandidatesSchamea"? I don't think "Saved" means anything here.

apps/yapms/src/lib/types/Region.ts
```diff
@@ -24,6 +24,12 @@ export const RegionSchema = z.object({
 	})
 });
 
+export const RegionCandidateSchema = z.object({
+	candidate: CandidateSchema,
+	count: z.number(),
+	margin: z.number()
+});
+
```
Fish1, line 32: The "SavedRegionCandidatesSchema" could be based on this.

Then the RegionSchema could use the "SavedRegionCandidatesSchema".

Because this is duplicated 3 time otherwise.

apps/yapms/src/lib/utils/luma.ts
```diff
@@ -6,6 +6,25 @@ function hexToRGB(hex: string) {
 	return { r, g, b };
 }
 
+function blendHexForLuma(hexes: string[]) {
```
Fish1: This could probably be renamed to just "blendHexColor" or something.

Also, I think it's okay to return a Hex color with the # at the beginning, as all the other functions here account for it.

Fish1 (changes requested):
LGTM. I think just a few things should be renamed for clarity.

apps/yapms/src/lib/types/Region.ts
Flizzyclone: I chose that since the only time this data structure is used is when we are loading candidates in from either the SVG or JSON.

apps/yapms/src/lib/types/Region.ts
Flizzyclone: I think this should get a broader refactor where we standardize the ways in which region info is loaded. We have 4(?) different formats it is loaded in right now. I can make them derived from each other with the following code, but it is still messy.

```
import { CandidateSchema } from './Candidate';
import { z } from 'zod';

export const RegionCandidateSchema = z.object({
	candidate: CandidateSchema,
	count: z.number().nonnegative(),
	margin: z.number().nonnegative()
});

export const RegionSchema = z.object({
	id: z.string(),
	shortName: z.string(),
	longName: z.string(),
	value: z.number(),
	permaVal: z.number(),
	disabled: z.boolean(),
	locked: z.boolean(),
	permaLocked: z.boolean(),
	candidates: RegionCandidateSchema.array(),
	nodes: z.object({
		region: z.custom<HTMLElement>((val) => val instanceof HTMLElement),
		button: z.custom<HTMLElement>((val) => val instanceof HTMLElement).nullable(),
		text: z.custom<HTMLElement>((val) => val instanceof HTMLElement).nullable()
	})
});

export const SavedRegionCandidateSchema = RegionCandidateSchema.omit({ candidate: true }).extend({
	candidate:z.string()
})

export const SavedRegionCandidatesSchema = SavedRegionCandidateSchema.array()

export const SavedRegionSchema = RegionSchema.omit({
	shortName: true,
	longName: true,
	nodes: true,
	candidates: true
}).extend({
	candidates: SavedRegionCandidateSchema.omit({ candidate: true }).extend({
		id:z.string()
	}).array()
});

type Region = z.infer<typeof RegionSchema>;

export type RegionCandidate = z.infer<typeof RegionCandidateSchema>;

export type SavedRegionCandidates = z.infer<typeof SavedRegionCandidatesSchema>;

export default Region;
```

Fish1: Oh sorry gottcha, I wasn't paying attention to the fact that the SavedRegionCandidatesSchema represented the ID of the candidate in a file.

If you want you could just do this to remove the duplicates. Otherwise it's fine.

```ts
export const RegionCandidateSchema = z.object({
	candidate: CandidateSchema,
	count: z.number(),
	margin: z.number()
});

export const RegionSchema = z.object({
	id: z.string(),
	shortName: z.string(),
	longName: z.string(),
	value: z.number(),
	permaVal: z.number(),
	disabled: z.boolean(),
	locked: z.boolean(),
	permaLocked: z.boolean(),
	candidates: RegionCandidateSchema.array(),
	nodes: z.object({
		region: z.custom<HTMLElement>((val) => val instanceof HTMLElement),
		button: z.custom<HTMLElement>((val) => val instanceof HTMLElement).nullable(),
		text: z.custom<HTMLElement>((val) => val instanceof HTMLElement).nullable()
	})
});

export const SavedRegionCandidatesSchema = z
	.object({
		candidate: z.string(),
		count: z.number().nonnegative(),
		margin: z.number().nonnegative()
	})
	.array();

export const SavedRegionSchema = RegionSchema.omit({
	shortName: true,
	longName: true,
	nodes: true,
	candidates: true
}).extend({
	candidates: SavedRegionCandidatesSchema
});
```

Flizzyclone: This breaks loading saved maps (they use id: instead of candidate: ) so I will leave it out and we can refactor at another time.

Fish1 (approved):
LGTM

PR #347, opened 2023-09-09
USA Counties 2020 map

Flizzyclone:
This PR adds a USA counties 2020 map (connecticut changed county equivalents between 2020 & 2023)

sourcing included in the svg. 2023 counties also gets sourcing.


Fish1 (changes requested):
The counties have card coded styles for the stroke-width. I think they just need to be removed.

image

Flizzyclone:
Good catch. Fixed.

Fish1 (commented):
LGTM

Fish1 (approved):

PR #349, opened 2023-09-10
Chart Leans

### LINKED ISSUE #348: Chart Leans Options
The user should have the ability to show or not show leans on every chart option.

Fish1:
Feature: Added the ability to enable & disable chart leans for all charts.

Code: Added a store to track if the user wants to see chart leans. Then add calculations in the charts to calculate both with and without leans.

image
image
image
image


Flizzyclone (approved):
LGTM. Wish we could make it so we weren't calculating two arrays, but they aren't reactive to region changes if we use functions.

PR #352, opened 2023-09-10
Added apple-touch-icons

Fish1:
I noticed that there were lots of warning logs about apple icons. So I just added them.


Flizzyclone (approved):
This looks good- do we need a favicon.ico as well? I remember seeing warnings for that.

PR #353, opened 2023-09-11
Region tooltip

Flizzyclone:
This PR adds a tooltip to display a regions short-name on hover. This starts off disabled but can be enabled in the options modal.
image
yapmsBasicTooltip

Summary of Code Changes:
- Adds @floating-ui/dom dependency
- Adds option to toggle the tooltip
- Creates component for the tooltip
  - Position is determined by floating-ui, with the cursor position passed to RegionTooltipStore as input.
  - The tooltip shows under the condition that 400ms has passed w/o mouse movement & the cursor is on a region
- Adds RegionTooltipStore
- Adds mouse listeners to update RegionTooltipStore

### DIFF
--- a/apps/yapms/package.json
+++ b/apps/yapms/package.json
@@ -18,6 +18,7 @@
 		"format": "prettier --plugin prettier-plugin-svelte --write ."
 	},
 	"devDependencies": {
+		"@floating-ui/dom": "^1.5.1",
 		"@fontsource/roboto": "^5.0.8",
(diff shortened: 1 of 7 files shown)


Fish1 (changes requested):
I think one performance thing too make here would be to delete the unused timeouts, so it doesn't need to update the store more than needed. You can do that by storing them in a variable and deleting them before you make the next one.

This should prevent a bunch of store changes. Not too sure what the performance differences are though.


image


image

Fish1 (approved):

PR #359, opened 2023-09-12
Client side screenshot

### LINKED ISSUE #314: Client Side Screenshots
There should be a client side screenshot feature, so people can copy an paste images without needing to manually do it.

Fish1:
Feature: On the share modal, the user can take a screenshot that will download to their device.

Code: Added htmlToImage, passed the element to the screenshot function and save it using fileSaver.

### DIFF
--- a/apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
+++ b/apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
@@ -109,13 +109,15 @@
 			return;
 		}
 
-		const data = await htmlToImage.toBlob(mapChart, {
-			backgroundColor: 'hsl(var(--b1))'
-		});
+		mapChart.style.backgroundColor = 'hsl(var(--b1))';
+
+		const data = await htmlToImage.toBlob(mapChart);
 
 		if (data !== null) {
-			fileSaver.saveAs(data, 'screenshot.png');
+			fileSaver.saveAs(data, 'YAPmsScreenshot.png');
 		}
+
+		mapChart.style.backgroundColor = '';
 	}
 </script>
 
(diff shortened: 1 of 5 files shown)


apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
```diff
@@ -99,13 +102,32 @@
...
 	}
+
+	async function screenshot() {
+		const mapChart = document.getElementById('map-chart-div');
+		if (mapChart === null) {
+			return;
+		}
+
+		const data = await htmlToImage.toBlob(mapChart, {
+			backgroundColor: 'hsl(var(--b1))'
```
Flizzyclone: This does not work for me. Can we make map-chart-div have this as its background? Do we make backgrounds transparent?

apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
```diff
@@ -99,13 +102,32 @@
...
+		});
+
+		if (data !== null) {
+			fileSaver.saveAs(data, 'screenshot.png');
```
Flizzyclone: Nit: maybe we make this YAPmsScreenshot like YAPmsMap?

Flizzyclone (changes requested):

apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
Fish1: What does it look like for you?

apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
Fish1: Changed

apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
Fish1: I did your suggestion and it works on chrome now. Submitted a bug report with html-to-image to see if it can work on chrome.

Flizzyclone (approved):
Working for me now, LGTM

PR #365, opened 2023-09-17
Cleanup Glob Imports

Fish1:
We can actually use dynamic imports, as long as the $lib variable isn't used. This allows for some easier to read code.


Fish1:
Sorry, this was an accident.

PR #383, opened 2023-09-27
Side bar on mobile

Flizzyclone:
This PR adds support for the side bar to appear on mobile (starts minimized)

This is what it looks like on a Pixel 5 in DevTools:
image

If it doesn't need to (in this case there are no links), the side bar will not cover the whole screen:
image

Summary of code changes:
- Makes the side bar toggle button always appear
- Changes classes so that the side bar is positioned absolutely when screen width is less than md
- Changes default SideBarStore value to false
- Adds logic to change SideBarStore to true if screen width is greater than 768 (Tailwind's md breakpoint)


Fish1 (approved):
LGTM, I think one thing to make it easier, would be to make the sidebar button always visible.

PR #384, opened 2023-09-28
Allow Maps & Charts to shrink

Fish1:
Some maps overflowed on mobile devices. This should allow the map and chart to shrink/grow as needed.

Animation
Animation


Flizzyclone (approved):
LGTM

PR #385, opened 2023-09-30
Border Size Slider

Flizzyclone:
This PR adds a slider to control a multiplier for border size in the options menu
image

Summary of Code Changes:
- Adds input slider to OptionsModal
- Creates AutoStrokeMultiplierStore to store an int (default 1)
- Creates updateAutoStroke to force update the auto-border-stroke-width css property
- Changes auto border stroke width to be multiplied by AutoStrokeMultiplierStore value


Fish1 (approved):
LGTM

PR #400, opened 2023-10-13
Canadian provincial maps

Flizzyclone:
This PR adds Canadian provincial maps for every province minus Nunavut

I used the latest boundaries for each province. For all but Saskatchewan who has new boundaries for 2024, this meant that I used the last time an election was held for the year param.


Fish1 (approved):
LGTM

PR #411, opened 2023-10-21
Integrate titles into SVG

Fish1:
This will eliminate the need to maintain the titles.json file.

Titles & routes can be determines from the $lib/assets/map file structure.

Titles can be read from the SVGs.


apps/yapms/src/routes/+page.server.ts
```diff
@@ -1 +1,30 @@
...
+	const search = [];
+
+	for (const file of files) {
+		const data = fs.readFileSync(file, 'utf8');
+		const title = data.match(/title=['"](?<title>[^'"]*)['"]/)?.groups?.['title'];
+		if (title === undefined) {
+			continue;
+		}
+		const route = '/app/' + file.split('/').pop()?.split('.').at(0)?.replaceAll('-', '/');
+		search.push({
```
Flizzyclone: the file list is "backwards" so i find that replacing this search.push with a search.unshift gives us far better search results (e.g when you type in "2024 pres" it takes you to the normal 2024 presidential map instead of the take-all map

apps/yapms/src/routes/+page.server.ts
```diff
@@ -1 +1,30 @@
...
+import { globSync } from 'glob';
+import fs from 'fs';
+
+export const load: PageServerLoad = () => {
+	const files = globSync('./src/lib/assets/maps/**/*.svg');
+
+	const search = [];
+
+	for (const file of files) {
+		const data = fs.readFileSync(file, 'utf8');
```
Flizzyclone, line 11: i get an error during build related to using 'fs' here. 

"Failed to resolve entry for package "fs". The package may have incorrect main/module/exports specified in its package.json"

Flizzyclone (changes requested):
this looks mostly good code wise, but i hit errors on build. one i hit is described below because i could track it down, but i also hit this one:
image

apps/yapms/src/routes/+page.server.ts
Fish1: What command are you using to build?

Flizzyclone: npm run build

apps/yapms/src/routes/+page.server.ts
Fish1: Nice catch!

apps/yapms/src/routes/+page.server.ts
Fish1: I'm gonna guess this had something to do with being on windows vs linux.

Fish1:
Issues should be resolved now.

Flizzyclone (approved):
LGTM

PR #412, opened 2023-10-22
Remove map selection json ** REVIEW AFTER PR 411 **

Fish1:
** REVIEW AFTER https://github.com/yapms/yapms/pull/411 **

This is to remove the JSON for all the modals and buttons on the homepage. The cards are now components. Each card will contain a list of links to render.


Flizzyclone (approved):
fixed two card things (see my commit). LGTM

PR #414, opened 2023-10-25
Fix search for a Windows server

Flizzyclone:
This PR fixes a problem where if you are running the server on Windows, the separator for paths will be wrong & as a result, you will get redirected to some bad path from search results.


Fish1 (approved):
Works for me!

PR #415, opened 2023-10-26
New North Carolina maps

Flizzyclone:
Yesterday, North Carolina adopted a new Congressional map necessitating an update to the 2024 house map.

They also adopted new state house and state senate maps that this PR also adds


Fish1 (approved):
LGTM

PR #421, opened 2023-10-28
Handle the yapms.com url

Fish1:


Flizzyclone (approved):
ooh boy. lets do this

PR #428, opened 2023-10-30
SAF maps & import fixes

Flizzyclone:
This PR adds 4 South African maps
- 2019 National Assembly (blank, results)
- Provinces
- 2023 Municipalities

This also fixes three bugs w/ importing, two created from recent format change, one longstanding
- regions attribute being set instead of map-type
- region attribute being sanitized out (and thus all regions having the same undefined id)
- for-region attribute being sanitized out (longstanding, we never adequately updated that)


Fish1 (approved):
LGTM

PR #432, opened 2023-11-01
Always update sources on sidebar

### LINKED ISSUE #424: Sources Get Stuck
The sources list gets stuck between pages because it's in a global store.

Fish1:
This should make it so the source list always updates when the user open a new map.

closes #424


Flizzyclone (approved):
LGTM. good refactor as well.

PR #434, opened 2023-11-02
Navigate Home Modal

Fish1:
This is a popup asking if the user really wants to navigate to the home page, because it will result in a loss of their progress.


Flizzyclone (approved):
LGTM

PR #435, opened 2023-11-02
Fix South African Card & Country Code

Flizzyclone:
When South African maps were changed to Julian date, the card wasn't updated. This PR fixes that along with changing all routes from country code "saf" to the official country code "zaf".


Fish1 (approved):
LGTM

PR #440, opened 2023-11-04
World countries map

Flizzyclone:
This PR adds a map for all of the world's countries. For the purposes of this map, a country is defined (imperfectly) as a country that is a UN member state, General assembly observer, or recognized non-member state. An exception is made for Kosovo.

This PR also makes adding grid-cols-2 to MapCard's dependent on links length being greater than 2. This makes single or double links horizontally centered within the card instead of at the top left.

A three letter country code does not exist for the world so I used an abbreviated version of the world globe.


Fish1 (approved):
LGTM. Just made it so its horizontal when there are two links.

PR #441, opened 2023-11-04
Fix SEO issues pointed out by lighthouse on homepage.

Fish1:


Flizzyclone (approved):
changed a little bit of text for the open svg header, says "Open from SVG following YAPms format" now.

PR #443, opened 2023-11-04
Fix zooming on click for piechart

Fish1:
When using the pie chart and the user updates the map, the map would then center. This removes the recenter when the piechart gets rendered.


Flizzyclone (approved):
LGTM

PR #445, opened 2023-11-04
Replace Aqua with Dracula

Fish1:
I thought 3 light mode and 3 dark mode themes would be good.


Flizzyclone (approved):
fixed color order in the UI, besides that LGTM

PR #460, opened 2023-11-05
Disable the save button on imported maps.

Fish1:
The user shouldn't be able to save to their profile when they are viewing an imported map.

### DIFF
--- a/apps/yapms/src/lib/components/sidebar/sections/savedmaps/SaveMap.svelte
+++ b/apps/yapms/src/lib/components/sidebar/sections/savedmaps/SaveMap.svelte
@@ -2,7 +2,7 @@
 	import { PocketBaseStore } from '$lib/stores/PocketBase';
 	import { generateJson } from '$lib/utils/saveMap';
 	import PlusCircle from '$lib/icons/PlusCircle.svelte';
-	import { ImportedSVGStore } from '$lib/stores/ImportedSVG';
+	import { page } from '$app/stores';
 
 	export let disabled = false;
 	export let onSubmitted: () => void;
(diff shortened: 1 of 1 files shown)


apps/yapms/src/lib/components/sidebar/sections/savedmaps/SaveMap.svelte
```diff
@@ -37,13 +38,13 @@
...
 		placeholder="Map Name"
-		disabled={disabled || submitting}
+		disabled={disabled || submitting || $ImportedSVGStore.loaded}
 	/>
 	<div class="tooltip" data-tip="Create">
 		<button
 			class="btn btn-sm btn-primary join-item"
 			on:click={createMap}
-			disabled={disabled || submitting}
+			disabled={disabled || submitting || $ImportedSVGStore.loaded}
```
Flizzyclone: this condition is bad- $ImportedSVGStore.loaded will still have content if you import a map and then leave that page and navigate to another.

Flizzyclone: do we lock it based on path??

Flizzyclone (changes requested):

apps/yapms/src/lib/components/sidebar/sections/savedmaps/SaveMap.svelte
Fish1: I think the state should be changed to false if a map isn't loaded.

Flizzyclone: im cool w/ that, if you wanna have this PR manage that and then use this condition that is good

Fish1: Eh, your idea looks more clear in the code I think.

Flizzyclone (approved):

PR #468, opened 2023-11-06
Remove minimum zoom level

Flizzyclone:
This PR fixes an issue where the min zoom level would be too high for some maps and you couldnt zoom out to fit the entire map on your screen. Now you can zoom out infinitely.


Fish1 (approved):
LGTM

PR #469, opened 2023-11-07
Action Groups

Fish1:
This adds action groups to the SVG and to each path.

<svg action-groups="States, Regions">

<path action-groups="1, 1" />
<path action-groups="1, 2" />

For example...
This SVG has two groupings. States and Regions.

If you select the States group.
The first path is group one of States.
The second path is group one of States.
So when you click on state one or state two, they both get the action done on them.

If you select the the Regions group.
Then the first path has group 1
And the second path has group 2.
So only one of them will get the action done on them.

[Screencast from 2023-11-06 22-12-27.webm](https://github.com/yapms/yapms/assets/13600696/5bf4cbc5-b935-4ef7-b7b9-1e18b00dc5d0)


Flizzyclone (approved):
LGTM

PR #470, opened 2023-11-07
USA States map & primary updates

Flizzyclone:
This PR adds a blank USA States map and updates the candidates for the primaries to reflect who has declared and who has dropped out.


Fish1 (approved):
LGTM, just made the stroke match the presidential map, and renamed it to states_and_territories

PR #476, opened 2023-11-11
Cleanup Images

Fish1:
Cleanup the images folder, delete unused svgs, and put images inside the images folder.

Also added enhanced-img. This will convert pngs to modern formats like webp and avif if the clients web browser supports it.


Flizzyclone (changes requested):
enhanced images is breaking custom logos for me- I cannot upload one without getting an error in the console and the options surrounding chart logo breaking

my best guess is this is because whatever is uploaded isn't "enhanced" like it needs and the logos are because you are importing with ?enhanced at the end,

Fish1:
Eh, it's not really needed anyways. I've removed it.

PR #478, opened 2023-11-14
Fix New Zealand titles & 1964 pres

Flizzyclone:
This PR fixes two map bugs that have been reported:
- New Zealand had titles flipped for two maps, this fixes that.
- The USA 1964 presidential had a display issue where there would be random CSS text behind it


Fish1 (approved):
LGTM

PR #480, opened 2023-11-15
Run Screenshot on a GoRoutine

Fish1:
This will allow the user to get the link to their map while their screenshot is still processing.


Flizzyclone (approved):
LGTM

PR #500, opened 2023-11-18
Alaskan county fix

Flizzyclone:
This PR fixes a problem with an Alaskan county that split in 2019 but wasn't split on our maps.


Fish1 (approved):
LGTM. I just added the action groups tag.

PR #510, opened 2023-11-26
Improved USA counties

Flizzyclone:
This PR replaces the USA counties map with a new, higher quality version of it that also fixes outstanding issues with the original map. These maps are completely backwards compatible with saves of the original versions, though some regions (4 small virginia cities) not represented on the original will be unfilled.


Fish1 (approved):
LGTM

PR #523, opened 2023-12-02
Fix counties (for real)! and add dev warning for duplicate region IDs

### LINKED ISSUE #522: Virginia's counties don't work.
When I click on a county in Virginia, a county in Maryland changes instead of the one I'm clicking on.

Flizzyclone:
This PR fixes the last problem with the new counties map and adds a warning on region load for duplicate region IDs so we can hopefully avoid and at least easily debug this in the future. Checking for duplicates runs only in a dev environment to avoid running costly array logic every time a map is loaded.


Fish1 (changes requested):
Little bit of a nit pick, since this code only runs in dev.

But could you use a map instead of an array here? For constant time lookup.

Flizzyclone:
good point, will fix in a bit here

Flizzyclone:
oh fixes #522 btw

Fish1 (approved):
LGTM

PR #528, opened 2023-12-05
Presidential Take-All with Territories

Flizzyclone:
This map combines the USA take-all map (no split Nebraska, Maine) with the Presidential map with Territories represented

This map comes as the result of a request from someone on Discord who has an alt history where this would be useful.


Fish1 (approved):
LGTM

PR #533, opened 2023-12-18
Different projections for importing maps

### LINKED ISSUE #531: Advanced import option: projection specification using projection string
This feature would allow a user to change the projection their map is projected in by specifying a projection string

This would have to use proj4js (i read something the other day about this and actually wrote code to do this that i discarded, ugh!)

see here: https://gist.github.com/armollica/e6371dcceed6d89312cf6a78be1a0b49

use the import options dropdown from https://github.com/flizzyclone/yapms/tree/statesfun

Flizzyclone:
This PR adds the ability to import maps using different projections. For instance, someone can now project American maps in Albers USA projection like Wikipedia does and project Canadian maps the same way the statistics service does.

https://github.com/yapms/yapms/assets/42476312/bd182967-3270-4181-8d07-1ef54fb8c388

Summary of Code Changes:
- Adds proj4 as a dependency
- Rewrites ImportModal functionality to use a generic higher order function for loading
- Adds an Import Options dropdown in ImportModal to change projections
- Adds a couple different projection options and a custom projection option
  - Projection options are defined with a string label and then a function that returns a GeoProjection object. This function is called when converting GeoJSON to SVG to get a projection
- Adds proj4ToProjection which converts a proj4 string to a usable projection
- Adds two new fields to ImportedSVGStore under options
  - projectionFunction stores the function that gives a GeoProjection object
  - customProjectionDefinition stores the proj4 string the user enters


Fish1 (approved):
LGTM

PR #534, opened 2023-12-19
Added Donut Chart Type

Fish1:
The donut chart is the same as the pie chart. All we should need to do is pass in the 'doughnut' string.


Flizzyclone (approved):
LGTM

PR #536, opened 2023-12-19
Make Candidate Buttons Bigger

Fish1:
Makes the button font a bit bigger by default, also give the user the option to set the size to their liking.


Flizzyclone (approved):
LGTM

PR #549, opened 2023-12-29
Revised Georgia (US) maps

Flizzyclone:
This PR updates the USA house map & adds new maps to reflect the state of Georgia passing new district boundaries and having those boundaries upheld in court. Symlink for USA house 2024 is updated & title for the old map has been updated to reflect that it is an old version reflecting boundaries on October 25th.


Fish1 (approved):
LGTM

PR #580, opened 2024-02-09
Ability to map imported region properties to GeoJSON properties

Flizzyclone:
This PR adds a few new settings to the import menu that allow you to specify properties in GeoJSON for import functionality to use when importing a map. This won't work with shapefiles.

this is quite in the weeds


apps/yapms/src/lib/components/modals/importmodal/ImportOptions.svelte
```diff
@@ -59,6 +65,58 @@
 					</label>
 				{/if}
 			</div>
+			<span class="divider m-2"></span>
+			<div class="flex gap-x-1">
+				Property Settings (Advanced)
```
Fish1: I think this should be changed to "GeoJSON Properties" or something. Just to make it a bit more clear that these are setting for GeoJSONs.

Fish1 (changes requested):

Fish1 (approved):
LGTM

PR #581, opened 2024-02-10
Handle NaN on map load

Fish1:
If the map loader cannot parse a number it creates a NaN value, which the rest of the program cannot handle.


Flizzyclone (approved):
good bugfix

PR #589, opened 2024-02-16
Toggle Texts

Flizzyclone:
This PR adds a new option to toggle region texts, primarily for USA maps.

To make this setting also hide all region buttons for the USA maps, I went through and re-did all maps that use buttons to organize those regions into groups. When hiding texts, it will also hide all paths with the "button" class. So, group structure has one path without the "button" class (the state itself) and then one with the "button" class (the button, woah!). So, only the button gets hidden.


apps/yapms/src/routes/app/[country]/+layout.svelte
```diff
@@ -46,6 +47,7 @@
 		id="map-div"
 		class="overflow-hidden h-full"
 		class:insets-hidden={$MapInsetsStore.hidden}
+		class:texts-hidden={!$RegionTextsStore.enabled}
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/+layout.svelte]: Could you reverse the logic, so it's an option to hide the RegionTextStore? To match the MapInsetStore.hidden option.

Fish1 (changes requested):

Fish1 (approved):
LGTM, maybe in the future we can get rid of all the duplicated SVGs so changes are easier.

PR #595, opened 2024-02-22
Ohio 2024 maps

Flizzyclone:
Ohio updated its 2024 maps a while ago and someone brought it to my attention so this PR adds these new maps.

It also fixes a tiny bug I found in import.


Fish1 (approved):
LGTM

PR #596, opened 2024-02-22
Expose "default value" candidate property to users

Flizzyclone:
This PR adds a field in the Edit Candidate modal to change the starting value for a candidate (aka the value whey have when no regions are assigned).

I had to make the CandidateCounts and CandidateCountsMargin stores derive from CandidatesStore as well so they would properly update on a change of a candidate's default value.

yapmsStartingValue


Fish1 (approved):
LGTM

PR #599, opened 2024-02-26
Fixes to can-qc-2022-blank.svg

Flizzyclone:
A friend was using the Quebec map and found some issues. This PR fixes them.

The map had:
- Two regions assigned to a candidate
- A missing short-name/long-name
- No title


Fish1 (approved):
LGTM

PR #607, opened 2024-03-07
Update USA Dem candidates

Flizzyclone:
Two candidates in the USA Dem primaries that were not there before have gotten delegates, Jason Palmer and Uncommitted (not technically a candidate but you know). 

This PR adds them to the primary map


Fish1 (approved):
LGTM, just moved uncommitted to the end of the list.

PR #608, opened 2024-03-07
Fix 1836 candidate name

Flizzyclone:
1836 presidential results had the Democratic nominee as Franklin Pierce, not the real nominee, Martin Van Buren.


Fish1 (approved):
LGTM

PR #611, opened 2024-03-20
Mexico 2024 Legislative election maps

### LINKED ISSUE #610: Mexico 2024 Legislative election maps
Talked to @Flizzyclone about adding these maps so I started extracting the federal electoral districts from a National Electoral Institute (INE) map.

A few questions:
- There are 500 legislators in the Chamber of Deputies: 300 by direct vote (i.e. one per district) and 200 by proportional representation (40 per [circunscription](https://en.wikipedia.org/wiki/Electoral_regions_of_Mexico)). Given that the latter 200 depend on total vote percentages, I'm considering making the map such that it only accounts for the 300 directly elected deputies. How should this be handled?
- Same question with the Senate, 32 out of the 128 senators are PR (no circunscriptions, just total vote percentages).
- The remaining 96 senators are split into 2 groups: 64 relative majority and 32 first minority. There are 32 federal entities, whoever comes first place gets 2 seats (relative majority) and second place gets 1 (first minority), therefore, there's always 2 different parties for each state (and sometimes the 2 relative majority may come from different parties through coalitions, so 3 different parties). I thought this could be represented by having a map of the states and two dots inside each state (or off the coast if the entity is too small). Would this approach be right or would you suggest something else?

hanrodz:
- Chamber of Deputies
    - map of 300 districts worth 1 vote
    - map of 5 circunscriptions worth 40 votes
- Senate
    - map of 32 federal entities with 2 votes
    - one dot per federal unit worth 1 vote
    - proportional representation box worth 32 votes

Note: I did not test this, couldn't get a container engine to work on Apple Silicon. :P

Closes #610


Flizzyclone:
Alright so just a couple issues that I'm fixing on my end because I have a working instance:
On the Senate map, `for-region="Proportional"` should be `for-region="PR"` (It goes by region ID)
I scaled the Chamber of Deputies map down from being like 11000x8000. It causes some problem I can't remember when the maps are super big.

Conventions:
Removed the padding on the edges of the map to fit w/ convention.
We use julian dates for the date parameter on maps in case they change. This one shouldn't, but we keep it consistent. Going to set the date on these to 2024001.

And then one nitpick that I can act on if you are cool with it:
- The proportional text was just big, I scaled it to 75% and it looks good

Anyways these are absolutely fantastic. Adapting this from a pdf is a cool and novel way to do this, smart way to work around having no shapefile. Attributes are assigned 100% correct and format looks great. Thank you bunches for getting these made.

License on the image used looks good.

I will make the changes above, compress these maps & image and then if Fish approves one of us will merge

I feel like a big scary maintainer saying all this lmfao

Fish1:
Thanks for making this, I could never find a working Mexico map.

What tool did you use to make these maps?

As for testing... the containers are only for running some of the back-end services. You only need NPM installed locally if you want to run yapms frontend. And you can run the yapms server locally with just go installed.

1) The text on the proportional button for the senate is blocking the button from being clicked.

2) The source links should point to the web page that the maps were sourced from, a webpage with a button to the download hopefully.

3) The chamber of deputies map is pushing the side bar off the screen, even when the sidebar is supposed to be open. I'm assuming this is to do with viewport stuff.
[Screencast from 2024-03-19 21-51-00.webm](https://github.com/yapms/yapms/assets/13600696/4378d983-b261-43b9-a203-0517f4b932af)

4) Could you just remove the 2024 from the chamber of deputies text? Until we find a solution to fit more text onto the buttons.
image

5) Could you compress the Mexico image to something under 20kb? (as compressed as possible without making the image distorted on the homepage. Right now its 707KB. You can use googles webp cli tool for that if you'd like.
https://developers.google.com/speed/webp/docs/cwebp
image

6) Most of the map looks good (I'm assuming you used some sort of raster to svg tool to do this?) But there is one part of the map that should probably be fixed. Veracruz de Ignacio de la Llave on the senate map has this circle inside of itself.
image


Maybe for the future... but the deputies map is about 1MB now, it can probably be smaller. There are a lot of duplicated points in the map. Probably a result of the tool you used. Things like this could be straight lines instead.
image

hanrodz:
Thank you two for the feedback.

> What tool did you use to make these maps?

@Fish1 I retrieved the [circunscription maps](https://cartografia.ife.org.mx/sige7/?mapoteca=planos&mrm), turned into SVGs with Inkscape, removed all <path> objects with stroke color different from `#808080`, and then manually joined the borders of each district. Joined them by state and circunscription for the other maps.

**Addressing feedback from @Flizzyclone**
1. Fixed the `for-region` attribute
2. I'm not sure I understand the date format but the general election is happening on June 2, 2024, in case you wanna change it to that instead of 2024001.

**Addressing feedback from @Fish1**
1. See point 1 of above section
4. Included link to INE/IFE website
5. Assuming the sidebar issue will be solved with the rescaling @Flizzyclone will do
6. Would "Chamber of Deputies '24" be okay? I feel like it could be misleading to remove it given that districts are different from 2021 ones
7. Downsized the image to a little under 20KB
8. Fixed Veracruz de Ignacio de la Llave in both maps

Fish1:
I tested "Chamber of Deputies '24" and it seems to fit. So that should be fine.

Also, could you change the titles of the maps to be the country adjective? Just to match the other maps. Should be "Mexican"
image

hanrodz:
Done, shortened the çard name and changed both maps names to Mexican. :)

Flizzyclone:
> I'm not sure I understand the date format but the general election is happening on June 2, 2024, in case you wanna change it to that instead of 2024001.

We use the date on which the map went into effect instead of when the election is, basically. This system is designed around accommodating US house maps, which can change multiple times in one cycle. Do I wish it was more intuitive? Yeah.

> Assuming the sidebar issue will be solved with the rescaling @Flizzyclone will do

yup

Flizzyclone:
My changes:
- Resized to make deputies not huge and remove padding on maps to fit w/ convention.
- Changed to julian dating
- Optimized the maps with SVGO (file size reduction)

I played around with the maps a bit and all seems to be working great. I am approving.

Flizzyclone (approved):

Fish1:
Hold on, before this is merged.

I want to double check that these maps are not from a private company with copyright. Looking at the source again, it looks like this is someone else's work. I thought it was a government website.

Edit: oh just looked up the INE, that's fine.

Flizzyclone:
> Hold on, before this is merged.
> 
> I want to double check that these maps are not from a private company with copyright. Looking at the source again, it looks like this is someone else's work. I thought it was a government website.
> 
> Edit: oh just looked up the INE, that's fine.

Looks good to merge, then?

PR #612, opened 2024-03-22
Use gopngquant library

Fish1:
Swap out the pngquant package for a full golang package.

I built this myself using CGO. So there could totally be a memory leak somewhere...

Be prepared to revert back to old commit when this gets merged in.


Flizzyclone:
It should be noted that I had to install gcc to properly compile pocketbase under this commit (will that be an issue when we deploy?)

I am also getting this compilation error on my end, so I cannot properly test this.
image

PR #616, opened 2024-03-26
Japan House of Representatives

Fish1:


Flizzyclone:
Added more margins for the candidates to keep w/ convention, this looks fine to me

Flizzyclone (approved):

PR #622, opened 2024-04-14
Updated Wisconsin maps

Flizzyclone:
Wisconsin got new state-level maps in February, and GeoJSON was recently made available. This PR adds the new maps to YAPms


Fish1 (approved):
LGTM

PR #623, opened 2024-04-14
Enable SVG exporting for any map

Flizzyclone:
As requested on Discord, this PR allows exporting any map through the export SVG button in the Import modal.

Dunno why this wasn't enabled, I didn't even have to change the function for exporting at all.


Fish1 (approved):
LGTM

PR #632, opened 2024-04-23
Indian states/Lok Sabha

Flizzyclone:
This PR adds two Indian maps, one for their parliament, the Lok Sabha, and one for their states.

Only one odd things about this PR, the image is licensed under an Indian government specific license. Allows for commercial use just like a CC license would.


Fish1:
Could you cut the resolution in half to get around 20kb? It doesn't need to be 1200x600

Flizzyclone:
yeah changed to 1/2 res

Fish1 (approved):
LGTM

PR #639, opened 2024-05-11
1972-1990 US House maps

Flizzyclone:
This PR adds House maps for the US from 1972-1990, using the same source as the 1992-2014 maps (UCLA).


Fish1 (approved):
LGTM

PR #675, opened 2024-05-16
EditRegionModal input improvements

Flizzyclone:
This PR makes it so the EditRegionModal will submit upon pressing enter (preventNonNumericalInput was preventing this behavior) and so that the text input will be automatically focused upon the modal opening.


Fish1 (approved):
LGTM.

I'm not sure what the best practice is.

But another way to handle this would be to subscribe to the EditRegionModalStore. And you can react to stuff this way.

image

PR #676, opened 2024-05-19
Horizontal Bar Chart

Flizzyclone:
This PR adds the Horizontal Bar Chart from old YAPms. 

On the bottom:
image
On the left:
image

This line:
`style="grid-template-columns: {maxCandidateNameLength + 1}ch calc(100% - {maxCandidateNameLength +
		1}ch)"`
is particularly gross but I don't know of a better way to do it. Perhaps there is a good way with flex I don't know about to make sure the column for names is consistent size.

I tested this on both Firefox, Chrome and everything worked the same on both.


Fish1:
I think old yapms used the bar chart built into Chart.js , is there a reason you didn't use it here?

https://www.chartjs.org/docs/latest/charts/bar.html#horizontal-bar-chart

Flizzyclone:
Ahh that makes a lot more sense. No, there isn't really a reason I didn't use it here besides I didn't consider it. I'll see about reworking this later.

Flizzyclone:
closing for now

PR #677, opened 2024-05-20
Remove Content-Disposition header

Flizzyclone:
Safari (and iOS mobile browsers in general, it seems) have not been loading saved maps. This is because Safari was not properly decompressing the gzipped map data. [This issue](https://github.com/getsentry/self-hosted/issues/2285) suggested to remove the Content-Disposition header. I set it to empty and Safari worked.

Side note: I tried using the .Del function for the header instead of setting it to "" but that resulted in no change, either because the method doesn't work or (more likely imo), the header is set later in the process of sending the request if it is not already there.

Don't know why this started happening all of a sudden when we have presumably been sending this header forever.


Fish1 (changes requested):
It looks like pocketbase is setting it to attachment. 

I think when you set it to blank, it's just defaulting to "inline" so I think we should go with inline instead.

Fish1 (approved):

PR #678, opened 2024-06-02
Update broken Lok Sabha Region ID

Flizzyclone:
The Indian Lok Sabha map had a duplicate region ID. This fixes that


Fish1 (approved):
LGTM

PR #679, opened 2024-06-03
Border Color Option

Flizzyclone:
This PR creates a setting for changing the region border color to a couple different options. "Contrasting" makes use of the daisyUI color we use for insets. White and black are there as on some themes, the contrasting color is not that great.

I am not totally happy with calling another function on page load, but it's needed to define the CSS variable from cache. Setting the variable on document.body works without that function, but that seemed to be a worse approach imo.

### DIFF
--- a/apps/yapms/src/lib/stores/RegionStrokeColorStore.ts
+++ b/apps/yapms/src/lib/stores/RegionStrokeColorStore.ts
@@ -33,5 +33,5 @@ if (browser) {
 
 //This exists because map-div/SVG is not defined when the above code runs.
 export function setRegionStrokeColor(node: SVGElement) {
-	node.style.setProperty('--region-stroke-color', get(RegionStrokeColorStore));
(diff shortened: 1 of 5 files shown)


apps/yapms/src/lib/stores/RegionStrokeColorStore.ts
```diff
@@ -0,0 +1,37 @@
...
+		const mapSVG = document.getElementById('map-div')?.querySelector('svg');
+		if (mapSVG) {
+			mapSVG.style.setProperty('--region-stroke-color', colorsByKey[value]);
+		}
+	});
+}
+
+//This exists because map-div/SVG is not defined when the above code runs.
+export function setRegionStrokeColor(node: SVGElement) {
+	node.style.setProperty('--region-stroke-color', get(RegionStrokeColorStore));
```
Fish1: This needs to get the value from colorsByKey.

Fish1 (changes requested):

Fish1 (approved):
LGTM

PR #683, opened 2024-06-11
Apply auto border to rects

Flizzyclone:
The new border color setting exposed that auto borders were not applying to the DC label. This is because it is specified as a rect, not a path, and the CSS rule doesn't target it. This PR fixes that.


Fish1 (approved):
LGTM

PR #684, opened 2024-06-12
Key combo to recenter map

Flizzyclone:
This PR adds a key combination (shift+r) to recenter the map. Last two conditions are to make sure that the hotkey does not activate when the user is not focused on the map e.g typing in a text input.


Fish1 (changes requested):
Could this just be r instead of shift r?

Flizzyclone:
Yea, changed that

Fish1 (approved):
LGTM, the one issue I see is people might not have that map-div focused. And will wonder why it's not centering.

PR #686, opened 2024-06-15
Margins for split states

Flizzyclone:
This PR adds the ability to set margins for split states. Previously they could only take the first margin. Also updates pattern (stripe) generation to use the margins set.

https://github.com/yapms/yapms/assets/42476312/c1b18162-e27e-4e92-b570-07b7d89bcd8d


Fish1 (approved):
Awesome, glad this easily went into shared links. LGTM

PR #732, opened 2024-07-06
Switch Candidate Shortcut

Fish1:
Pressing 0-9 will switch between candidates, with 0 being the tossup candidate.

### DIFF
--- a/apps/yapms/src/routes/app/+layout.svelte
+++ b/apps/yapms/src/routes/app/+layout.svelte
@@ -30,11 +30,13 @@
 
 	function handleKeyDown(event: KeyboardEvent) {
 		$InteractionStore.set(event.code, true);
-		if (event.code === 'KeyR' && event.target instanceof Element && event.target.id === 'map-div') {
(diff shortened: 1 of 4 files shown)


apps/yapms/src/routes/app/+layout.svelte
```diff
@@ -32,6 +33,8 @@
 		if (event.code === 'KeyR' && event.target instanceof Element && event.target.id === 'map-div') {
 			reapplyPanZoom();
 		}
+
+		handleCandidateSelectionShortcut(event.code);
```
Flizzyclone: This should share the condition reapplyPanZoom() has that checks if the target is 'map-div'. Otherwise, pressing number keys anywhere on the page including while typing in an input will result in a candidate switch.

Flizzyclone (changes requested):

apps/yapms/src/routes/app/+layout.svelte
Fish1: Should be resolved now.

PR #854, opened 2024-09-02
"Tools" section

Flizzyclone:
This PR creates a "Tools" modal in the top bar where we can put various useful bits of functionality. The first bit I am attaching with this PR is Mass Edit functionality that allows you to edit the value of every state at once. 

Potential additional functionality to put here can include the much sought after simulation feature (I have ideas on how to make that a little more interesting than it was in YAPms1) and a table view of regions that allows you to fill, edit, etc. that you can sort alphabetically for copying from wikitables or tabular/list forms of data.


Fish1 (changes requested):
LGTM, if you could just put the tools button to the left of the login button.

Fish1 (approved):

PR #930, opened 2024-10-13
Alphabetical Homepage

Fish1:
Kept USA on top, because it's by far the most popular.


Flizzyclone:
LGTM. I initially had worries with putting CAN and GBR further down but they make up a tiny percent of usage compared to what I thought

Flizzyclone (approved):

PR #957, opened 2024-10-23
Britain 2024 Results

Flizzyclone:
This map has been requested a few times and was easy enough to make. Also updates the map card on the homepage.


Fish1 (changes requested):
The SVG needs all the proper tags for the sharing to work.

Such as country, year, type and variant.

Flizzyclone:
Fixed, good catch

Fish1 (approved):

PR #1021, opened 2024-11-22
Post-2024 USA updates

Flizzyclone:
This PR contains a couple of maps:
2024 Presidential, Senate, and Gubernatorial Results (I will make another PR for the house once everything is called)
2026 Senate, Gubernatorial, and House maps all symlinked to svg files with date 2024312 so they can be updated later if anything changes
2028 Presidential maps (just updated names)

And then it reworks the map cards accordingly


Fish1:
Nice work!

What's the reasoning to not use the same symbolic link pattern on the presidential that you used for the other maps?

Flizzyclone:
My reasoning there was that the presidential map does not change barring a change in the states but I suppose it is *technically* possible a state gets added. So I'll set that up.

Flizzyclone:
> My reasoning there was that the presidential map does not change barring a change in the states but I suppose it is _technically_ possible a state gets added. So I'll set that up.

Furthermore, that'll just make the url look better if we do use the symlink there

Flizzyclone:
Alright, presidential map dates have been changed to 2024312 to reflect the date I made the maps and symlinked to 2028. Map card also updated accordingly.

Fish1 (approved):
LGTM

PR #1093, opened 2024-12-25
Australia 2025 House

Flizzyclone:
Australia is having elections this upcoming year, and they just finished changing their districts. This PR adds the new districts.

License for the source is CC BY 4.0


Fish1 (approved):
LGTM

PR #1139, opened 2025-02-11
Numeric Input for Split

Flizzyclone:
This PR adds a numerical input to the split modal for precision when working with larger values.

yapmsSplitNumeric

This PR makes use of a CSS feature, `field-sizing` that is not yet widely available (lacking safari, firefox). To get around that, it uses an `@support` directive for progressive enhancement that will apply styles to make the input resize with content (as seen above) if supported by the browser. If not, the input will be a fixed size of 6 characters. The fixed size is a bit wonky, but will be phased out as Firefox and Webkit adopt the new style.

image
_Split Modal on Firefox_

`preventNonNumericalInput` and `preventNonNumericalPaste` are copy-pasted from the Edit Region Modal.


bloomunreal:
awesome sauce

Fish1 (approved):
LGTM.

In tailwind 4, there is a field size property that we could use when we update to it.

https://tailwindcss.com/docs/field-sizing

PR #329, opened 2023-09-05
Revert to checkbox for modals

Flizzyclone:
This PR changes modals from being dialog based back to being checkbox based. This prevents UI from freezing on Mac for modals that have more than just the open prop in their store. Tested on an M2 iMac.

Closes #323

Issue #323, opened 2023-08-31
UI no longer active after component rename


yessenbayev:
Under the "President" view if I edit the name of the party, (for example, rename "Democrat" to "Democrats") the UI is no longer active.

From the log, it is due to an unhandled promise. I paste the JavaScript Console log below

[Error] Unhandled Promise Rejection: InvalidStateError: The object is in an invalid state.
	at (scheduler.1babeacc.js:1:9185)
	promiseReactionJob

Platform: Safari Version 16.2 (17614.3.7.1.7, 17614)

Flizzyclone:
for my reference: https://discord.com/channels/1038861619540795512/1038875719666765954/1147372940430676062

this is caused by some weird stuff related to closing modals using dialogs. seems like when we update a store associated w/ a modal, the modal tries to figure out if it should close or not and in the process goes through too many updates at once.

PR #1140, opened 2025-02-11
Improved Modal Functionality

Flizzyclone:
This PR is a clone of #287 which changed modals to take advantage of the accessibility improvements of the dialog component (better focusing, can close by clicking outside, can close using escape key)

This PR
- Removes open property & checkbox
- binds dialog element to dialog variable, calls showModal and close on it when the open variable in store changes

We reverted #287 in #329 to fix an issue (#323) where UI would freeze on Mac if a property on the modal store other than the open property was modified. Something (probably a dependency) has changed since then and I've tested this on a Mac running Safari 17.3, Chrome, and Firefox without issue. I also tested this on an old IPad and it was fine.

Additionally, if it was a Safari change, Safari 17.3 (the version I tested on) and later make up above 95% of Safari installations.

All that to say, I think it's fine to merge back in.

### DIFF
--- a/apps/yapms/src/lib/components/modals/ModalBase.svelte
+++ b/apps/yapms/src/lib/components/modals/ModalBase.svelte
@@ -16,20 +16,17 @@
 	}
 
 	let content: HTMLDivElement | undefined;
+	let dialog: HTMLDialogElement | undefined;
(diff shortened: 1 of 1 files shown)


Fish1 (approved):
LGTM

PR #1141, opened 2025-02-13
Fix browserless URL in deploy-yapms playbook

Flizzyclone:
`deploy-yapms.yml` never got updated to use Browserless v2 so I broke images yesterday when I deployed using it.


Fish1 (approved):
LGTM

PR #1142, opened 2025-02-14
Polish Maps

Flizzyclone:
This PR adds some maps for Poland for their first and second level administrative divisions and their parliamentary constituencies.

Source data is from the Polish Governments National Register of Boundaries which is "provided free of charge and can be used for any purpose"

Image for the map card is in the public domain.


Fish1 (approved):
LGTM

PR #1157, opened 2025-02-18
Fix hiding insets & labels on imported maps

Flizzyclone:
This PR fixes two issues that made insets/labels not hide on imported maps when the corresponding button in options was pressed.

- When changing the class name to hide insets, the name was not changed in the imported route.
- When adding code to hide region labels/buttons, the texts-hidden class conditional was never added to the imported route.


Fish1 (approved):
LGTM

PR #1159, opened 2025-03-05
Regions Table Tool

Flizzyclone:
Apologies for the big PR, but this wasn't able to go in piecemeal. I have left annotations throughout this PR.

This PR adds a new "Regions Table" tool into the "Tools" section. This table, per its name, let's you modify region attributes within a table instead of on the map. This is useful for scenarios in which you aren't familiar with the geography. I have personally been using it for Canadian predictions.

In the table, you can:
- Filter table entries
- Set the winning candidate for a region
- Cycle the winning candidate's margin
- Open the split modal
- Modify the region's value
- Disable a region
- Lock a region
image


apps/yapms/src/lib/components/modals/ModalBase.svelte
```diff
@@ -27,7 +29,7 @@
 </script>
 
 <dialog class="modal modal-bottom lg:modal-middle" bind:this={dialog} on:close={close}>
-	<div class="modal-box flex flex-col w-full">
+	<div class="modal-box flex flex-col w-full" class:!max-w-full={fullWidth}>
```
Flizzyclone, line 32: !max-w-full overrides other max-height properties and grows the modal to take up all available screen width.

apps/yapms/src/lib/components/modals/toolsmodal/tools/TableModal.svelte
```diff
@@ -0,0 +1,128 @@
...
+				<thead>
+					<tr class="border-neutral">
+						<th style="width:15%;">Region</th>
+						<th style="width:20%;">Candidate</th>
+						<th style="width:10%;">Value</th>
+						<th style="width:5%;">Disabled</th>
+						<th style="width:5%;">Locked</th>
+					</tr>
+				</thead>
+				{#if $TableModalStore.open}
```
Flizzyclone, line 60: Here I prevent svelte from rendering table rows if the user is not going to use the table. The rows then load on navigation to the table.

 If this check isn't here, loading larger maps is quite laggy.

Flizzyclone, line 78: test

apps/yapms/src/lib/components/modals/toolsmodal/tools/TableModal.svelte
```diff
@@ -0,0 +1,128 @@
...
+	) {
+		const newValue = Number(event.currentTarget.value);
+		const index = $RegionsStore.findIndex((region) => region.id === regionToUpdate.id);
+		if (!$RegionsStore[index].disabled) {
+			//Don't update value if disabled so the state stays disabled!
+			$RegionsStore[index].value = newValue;
+		}
+		$RegionsStore[index].permaVal = newValue;
+	}
+
```
Flizzyclone: filterRegions below returns a map of the form (region id, boolean) where the bool determines whether or not to show that region in the table.

That is then used in this check:
`{#if !isFiltering || regionsToShow.get(region.id) == true}`

Where `!isFiltering` means all regions are shown if no filter is set.

apps/yapms/src/lib/components/modals/toolsmodal/tools/tablecomponents/CandidateActions.svelte
```diff
@@ -0,0 +1,135 @@
...
+			</select>
+		{:else}
+			<select
+				class="select select-bordered select-sm w-full grow"
+				value={getWinnerId(region)}
+				on:change={changeWinner}
+			>
+				{#if checkIfSplit(region)}
+					<option value="split">Split</option>
+				{/if}
```
Flizzyclone, line 80: This adds an extra "Split" option to the select if the region is split so that the user can still interact with the dropdown even if the region is split.

apps/yapms/src/lib/components/modals/toolsmodal/tools/tablecomponents/CandidateActions.svelte
```diff
@@ -0,0 +1,135 @@
...
+
+	function checkIfSplit(region: Region) {
+		for (const candidate of region.candidates) {
+			if (candidate.count == region.value) {
+				return false;
+			}
+		}
+		return true;
+	}
+
```
Flizzyclone, line 44: `getWinningColor` below just gets how the region is filled on the map.

There is an edge case here if the region is filled with stripes, as you must fill an HTML element with a linear-gradient instead and not an svg <pattern> (and vice versa). Hence, `makePatternCSS`.

apps/yapms/src/lib/utils/patterns.ts
```diff
@@ -35,3 +35,21 @@ export function makePattern(candidates: Array<RegionCandidate>) {
 	mapSVG?.appendChild(pattern);
 	return `url(#${name})`;
 }
+
```
Flizzyclone, line 38: This is the `makePatternCSS` function as noted above. This generates the same striped pattern used for split regions, but in CSS `linear-gradient` form so it can be used on HTML elements.

apps/yapms/src/lib/utils/patterns.ts
```diff
@@ -35,3 +35,21 @@ export function makePattern(candidates: Array<RegionCandidate>) {
...
+			`${winner.candidate.margins[winner.margin].color} ${colorInterval * i}% ${colorInterval * (i + 1)}%`
+		);
+	});
+
+	gradientString += colors.join(', ');
+
+	gradientString += ')';
+
+	return gradientString;
+}
```
Flizzyclone, line 55: This is the `makePatternCSS` function noted above. This generates the same striped pattern used for split regions, but in CSS `linear-gradient` form so it can be used on HTML elements.

apps/yapms/src/lib/components/modals/toolsmodal/tools/TableModal.svelte
```diff
@@ -0,0 +1,128 @@
...
+						<th style="width:10%;">Value</th>
+						<th style="width:5%;">Disabled</th>
+						<th style="width:5%;">Locked</th>
+					</tr>
+				</thead>
+				{#if $TableModalStore.open}
+					<tbody>
+						{#each sortRegions($RegionsStore) as region}
+							{#if !isFiltering || regionsToShow.get(region.id) == true}
+								<tr class="border-neutral">
```
Fish1: I think the border should be border-neutral-300 , and maybe the header border be border-neutral-400 , to make the lines between the rows a bit softer.

apps/yapms/src/lib/components/modals/toolsmodal/tools/tablecomponents/CandidateActions.svelte
```diff
@@ -0,0 +1,135 @@
...
+		} else {
+			candidateToBeEdited.margin =
+				currentCandidateMargin === 0
+					? candidateToBeEdited.candidate.margins.length - 1
+					: candidateToBeEdited.margin - 1;
+		}
+
+		region.candidates = [candidateToBeEdited];
+		$RegionsStore = $RegionsStore;
+	}
```
Fish1: I think this should be moved to /lib/stores/regions/regionActions.ts just to keep all the region manipulation logic in one spot. At least as much as possible.

I just copy & pasted the fillRegion function and added a variable for candidateID as a string. I think that can then be used in spot of your changeWinner and cylceWinnerMargin functions.

I think long term, the fillRegion function should be changed to take in a config object.

image

apps/yapms/src/lib/components/modals/toolsmodal/tools/TableModal.svelte
```diff
@@ -0,0 +1,128 @@
...
+	import { RegionsStore } from '$lib/stores/regions/Regions';
+	import type { Region } from '$lib/types/Region';
+	import ModalBase from '../../ModalBase.svelte';
+	import { disableRegion, lockRegion } from '$lib/stores/regions/regionActions';
+	import { get } from 'svelte/store';
+	import CandidateActions from './tablecomponents/CandidateActions.svelte';
+
+	let filterInput = '';
+	$: isFiltering = filterInput !== '';
+	$: regionsToShow = filterRegions(filterInput);
```
Fish1: optional nit-pick:

If you reduce the regionsToShow to this line of code, you can eliminate a lot of your logic. It will just create the sorted and filtered list of regions. And will automatically update as you type.

image

apps/yapms/src/lib/components/modals/toolsmodal/tools/tablecomponents/CandidateActions.svelte
Fish1: I was able to put the function here.

image

Fish1 (changes requested):
Looks good and works from my testing!

Just a small style tweak, and moving some of the functions to where the other region manipulation functions are.

apps/yapms/src/lib/components/modals/toolsmodal/tools/tablecomponents/CandidateActions.svelte
Flizzyclone: Right, this is good. It certainly felt as if we were lacking proper standardized ways to manipulate regions as I was doing this. I will move the logic for updateRegionValue as well.

apps/yapms/src/lib/components/modals/toolsmodal/tools/TableModal.svelte
Flizzyclone: This is still performant, so implemented.

apps/yapms/src/lib/components/modals/toolsmodal/tools/tablecomponents/CandidateActions.svelte
Flizzyclone: Did find this is kinda slow due to the region search. Note to self for if I come back to this: function that just consumes a region object

Flizzyclone:
Alright, this is ready for review again. I changed around the styling to make the header stand out more, the borders between rows are now consistent with what I was going for across all themes, and logic has been pulled out into regionActions.ts

Due to an update, the linter is throwing errors. None of them are about the files here. We will have to go through and fix them in another PR.

Fish1 (approved):
LGTM

PR #1185, opened 2025-03-23
Fix Linting Issues

Fish1:
This is to fix linting issues.

Turned off require-each-key to ignore errors on svelte loops.

Then removed unneeded reactive statements, and unused JavaScript in HTML.


Flizzyclone (approved):
LGTM

PR #1186, opened 2025-03-24
Update tailwindcss & daisyui

Fish1:
Upgraded to tailwindcss 4 and daisyui 5. 

Rewrote the options modal because the formatting was bad on many of the buttons and fields.


Flizzyclone (approved):
LGTM. Please review commits I made. Main change is adapting the import modal, since it was quite broke.

Flizzyclone:
Additionally, I noticed that the split region modal slightly changed. The tossup color is faded now as opposed to prod where it is the same color. It doesn't bother me, but I wanted to flag it and leave it up to you.

image

PR #1201, opened 2025-03-30
Update Modal Header Style

Fish1:
Made the headers/title of the modals match the bold styling of the fieldsets.

### DIFF
--- a/apps/yapms/src/lib/components/modals/ModalBase.svelte
+++ b/apps/yapms/src/lib/components/modals/ModalBase.svelte
@@ -34,11 +34,8 @@
 			{title ?? ''}
 		</h3>
 
-		<button
-			class="btn btn-sm lg:btn-xs btn-circle btn-error absolute right-4 top-4 lg:top-5"
-			on:click={close}
-		>
-			<XMark class="w-4 h-4" />
+		<button class="btn btn-md btn-circle btn-ghost absolute right-4 top-4" on:click={close}>
+			<XMark class="w-6 h-6" />
 		</button>
 
 		<slot name="header" />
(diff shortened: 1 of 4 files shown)


apps/yapms/src/lib/components/modals/ModalBase.svelte
```diff
@@ -30,25 +30,21 @@
...
-					<XMark class="w-5 h-5" />
-				</button>
-			</div>
-		</div>
+		<h3 class="font-bold">
+			{title ?? ''}
+		</h3>
+
+		<button
+			class="btn btn-sm lg:btn-xs btn-circle btn-error absolute right-4 top-4 lg:top-5"
```
Flizzyclone: Do you mind terribly if we keep the larger exit button? I know it looks goofy with the larger text, but it's also a lot easier to click.

apps/yapms/src/lib/components/modals/candidatemodal/EditCandidateModal.svelte
```diff
@@ -105,55 +105,57 @@
...
-						<MinusCircle class="w-6 h-6" />
-					{/if}
-				</button>
-			</li>
-		{/each}
-	</ul>
+<ModalBase title="Edit" store={EditCandidateModalStore} onClose={close}>
+	<div slot="content" class="flex flex-col gap-4">
+		<div class="flex flex-row gap-2 items-center w-full">
+			<input
```
Flizzyclone: Could we add labels to these other than just the placeholder? Otherwise, looks good.

Flizzyclone (changes requested):

PR #1202, opened 2025-03-30
Refresh Themes

Fish1:
Use different themes, just for fun.


Flizzyclone:
Added back other themes. Also moved Dark up to the spot it is in on prod.

Flizzyclone (approved):

PR #1205, opened 2025-03-30
Redesign Share Modal

Fish1:
Makes the share modal use field sets. Also splits up the local save features, and the link generation.

image


Flizzyclone (approved):
LGTM. I like this a lot better actually. Good clarification on what the Load/Save is for.

PR #1208, opened 2025-03-31
Map Modal Button Text Compaction

Fish1:
This will make the text on the home page buttons closer together, as they were nearly touching the edges of the buttons.


Flizzyclone (approved):
Took me a second to notice, but LGTM.

PR #1209, opened 2025-03-31
Fix sortable items

Flizzyclone:
This PR migrates us to using `svelte-sortable-items` instead of `sortablejs`.


Fish1 (changes requested):
If you could just lock down the svelte-sortable-items package to 1.0.1 .

Otherwise, LGTM.

Flizzyclone:
Done!

Fish1 (approved):

PR #1265, opened 2025-04-07
Fix IOS Zoom

Fish1:
This should prevent IOS from zooming in on text fields.

Found on...
https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone


Flizzyclone (approved):
Surprised we didn't have this tbh. LGTM

PR #1280, opened 2025-04-07
Return to sortable.js

### LINKED ISSUE #1272: Candidate Modal - svelte 5

Flizzyclone:
This PR returns us to using sortable.js. I found a hook-based adaptation for Svelte 5 in an article and used it.

This fixes the issue that you cannot drag on Firefox.

I also refactored these components into Svelte 5. As such, this closes #1272.


Fish1:
I think that same error still happens. I think the fact that it moves the DOM elements around before you drop it in the spot is the issue.

image

Fish1 (changes requested):
This still has the original issue of crashing the site.

Fish1 (approved):

PR #1397, opened 2025-05-29
Post-Election Canadian Updates

Flizzyclone:
This PR does 4 things regarding the Canadian House of Commons maps
- Makes slight fixes to Montreal boundaries
- Adds action groups for both provinces and insets
- Adds a 2025 Results map
- Updates the Canadian map cards accordingly


Fish1:
Why were the borders changed? Were they changed because they were changed in real life? Or were they simply fixed to match reality?

If we change it, it will change all the maps people have created.

Fish1 (approved):

PR #1405, opened 2025-05-31
Remove dotenv vault from build process

Fish1:
Uses public variables set in the GitHub settings to set public environment variables.


Flizzyclone:
How is the app getting secrets here? Previously we would have copied over the .env is my understanding. Are we keeping a .env file on the server somewhere?

Fish1:
> How is the app getting secrets here? Previously we would have copied over the .env is my understanding. Are we keeping a .env file on the server somewhere?

The frontend container doesn't need secrets, it just needs a few environment variables set, so I set them in ```settings > environment > production```

The backend container never got secrets from the .env file , the turnstile secret was sent via the ansible deployment script and the google & discord login secrets are set in the pocketbase UI.

Flizzyclone:
Oh duh, they're secrets. That makes sense.

Flizzyclone (approved):

PR #1421, opened 2025-06-05
Disable the Download button when using an imported map

Fish1:
Having the ability to download your map save file is confusing when using custom maps. It's better to just export the entire SVG and load that back in.


Flizzyclone (approved):
LGTM

PR #1430, opened 2025-06-07
Generate Shades and Candidate Modal Fixes

Flizzyclone:
This PR adds a new feature to generate margins automatically, exactly as I do when I create maps. The code is based upon https://github.com/hihayk/scale.

It also makes the following changes to the Add Candidate modal to standardize it with the Edit Candidate modal
- Adds ability to add default value
- Adds labels for name and default value inputs

The Edit Tossup modal also receives a minor style fix for Chrome:
image
*old*
image
*new*

And one last thing. I moved `preventNonNumericalInput` and `preventNonNumericalPaste` into a utility file so I don't keep copy pasting it everywhere. I will update other uses of those two functions to use the util version in another PR.


Fish1 (changes requested):
Logic seems good to me.

1) Could you lock the color package to the version you want, instead of having it float.

2) Could you abstract the shade generator into it's own component? It looks like most of the logic is just duplicated twice for both the Add and Edit modals.

Flizzyclone:
Done! Shades functionality is now in it's own component that takes a callback to update state in the parent. Package also locked to v5.0.0

Fish1 (approved):

PR #1440, opened 2025-06-11
TCT Game Import

Fish1:
Adds a TCT game load option in to the share menu. Maybe it should go under an "extra" drop down or something, not sure. It should set the states to the correct candidate, with the correct EC and disable it if it's not in the game.

image

Currently have a PR open on TCT to add in colors to their save files. 
https://github.com/newcampaigntrail/newcampaigntrail.github.io/pull/444

Two TCT save files for testing:
[tct_with_color.json](https://github.com/user-attachments/files/20740972/tct_with_color.json)

[tct_no_color.json](https://github.com/user-attachments/files/20740973/tct_no_color.json)

Also added in the neverthrow package, it's nicer than using try catch blocks IMO.


apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
```diff
@@ -43,6 +44,12 @@
 		}
 	}
 
+	function loadTCT() {
+		if (tctFiles && tctFiles.length > 0) {
+			loadFromTCTFile(tctFiles);
```
Flizzyclone: On success this should close the share modal like opening a normal file does.

apps/yapms/src/lib/utils/loadMap.ts
```diff
@@ -80,4 +113,140 @@ function loadFromJson(mapData: unknown): void {
...
 
-export { loadFromJson, loadFromFile };
+function convertTCTtoYapms(tct: unknown) {
+	const parsedData = FileSchema_TCT.safeParse(tct);
+	if (parsedData.success === false) {
+		return;
+	}
+
+	const yapmsData = {
+		map: {
```
Flizzyclone: Can we give some sort of feedback to the user if they try to import a map that is not a US map? For instance, this is a Netherlands map:
[tct_nederland.json](https://github.com/user-attachments/files/20746145/tct_nederland.json)

I looked into the schema and there doesn't seem to be a way to do this, which is fine, it's no big deal. Leaving this here still.

apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
```diff
@@ -159,6 +166,19 @@
 				</div>
 			</fieldset>
 
+			<fieldset class="fieldset flex flex-col gap-2">
+				<legend class="fieldset-legend">
```
Flizzyclone: Could you add "gap-1" to the class list here? That class makes the spacing between words follow the normal amount.

Flizzyclone (changes requested):
A couple things to note on UI/UX, but the core logic LGTM. Good work, this is a cool feature.

apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
Fish1: Added

apps/yapms/src/lib/components/modals/sharemodal/ShareModal.svelte
Fish1: Oh nice, didn't even notice.

apps/yapms/src/lib/utils/loadMap.ts
Fish1: Good idea, added a loop to verify that all the regions for the map exist in the list of states.

Flizzyclone (approved):
LGTM

Flizzyclone:
Oh one more thing I just noticed that is not a big deal but not sure if this is intended behavior: If I upload something like an SVG file here, it obviously doesn't work and doesn't redirect or anything which is good, but it also doesn't throw the "error reading this file" error.

Fish1:
> Oh one more thing I just noticed that is not a big deal but not sure if this is intended behavior: If I upload something like an SVG file here, it obviously doesn't work and doesn't redirect or anything which is good, but it also doesn't throw the "error reading this file" error.

Good catch, added error handling for that.

PR #1448, opened 2025-06-13
Flip Colors Feature

Flizzyclone:
This feature adds a button to quickly reverse the order of candidate colors as requested by a person on the Discord. Useful if people want to use a shading scheme opposite of what we normally do.

yapmsRotation2


Fish1 (changes requested):
LGTM, if you could just add the btn class to the reverse button, so it has feed back that would be good.

I think something like this keeps what you have, but gives it the button pop.
image

Fish1 (changes requested):
The add candidate modal needs the button class too.

Nitpick: I would recommend using the items-center class on the flex box, because it does the calculations to perfectly center everything, but the top margins look fine.

Flizzyclone:
Oops!

Changed to items-end, as items-center puts it slightly too far up.

image

Fish1 (approved):
LGTM

Sorry, yes. Ignore the nit pick, I was just reading the code of the button that didn't have the updated code.

PR #1459, opened 2025-06-18
Folders for Saved Maps

Flizzyclone:
This PR adds the ability to create folders with which you can organize your saved maps.

To add support for this on the backend, there are two migrations. The first one created the user_map_folders table with two properties, name and user. The second migration adds a folder property to the user_maps table.

### DIFF
--- a/apps/pocketbase/main.go
+++ b/apps/pocketbase/main.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"strings"
 
+	"github.com/pocketbase/dbx"
 	"github.com/pocketbase/pocketbase"
 	"github.com/pocketbase/pocketbase/apis"
 	"github.com/pocketbase/pocketbase/core"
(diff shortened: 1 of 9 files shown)


Fish1 (changes requested):
Could you also add a limit to how many folders someone can create, to prevent abuse. Maybe 100?

I'll also look into the same thing for the save files.

Flizzyclone:
Not acting on this rn because I want to have a more in depth discussion on how we wanna set limits. Will bring it up soon in web-development

Fish1:
> Not acting on this rn because I want to have a more in depth discussion on how we wanna set limits. Will bring it up soon in web-development

Okay!

Flizzyclone:
Okay. I am now limiting the number of folders and maps to 1000 using middleware, as we discussed. I also changed the way requests are made for folders. Instead of being made on page load, maps in folders are now only fetched when that folder is opened to prevent a bunch of requests on page load.

Fish1 (changes requested):
Doesn't seem to work for Firefox, at least on Linux.

Does it work for Firefox on windows for you?

Fish1 (changes requested):
Just a few cosmetic changes, but I think the logic is good.

* The new folder button should have some kind of style on it, I think the primary color looks fine.

* Folders should have an indicator on them to know they are folders when they are closed.

* Also the indicator should indicate if it's closed or open maybe.

* Maybe the loading animation should be moved or removed, because if it loads fast, it causes the animation to jitter a lot.

Fish1 (approved):
LGTM

PR #1466, opened 2025-06-20
Make save file navigate to URL for map.

Fish1:
This should make a save file navigate to the proper URL for a map.

This also reworks most of the map loading logic, as it seems to have been duplicated. This *should* unify the process.

### DIFF
--- a/apps/yapms/src/routes/app/[country]/+layout.svelte
+++ b/apps/yapms/src/routes/app/[country]/+layout.svelte
@@ -1,42 +1,58 @@
 <script lang="ts">
-	import { loadFromJson } from '$lib/utils/loadMap';
-	import { page } from '$app/stores';
 	import { MapInsetsStore } from '$lib/stores/MapInsetsStore';
 	import { applyAutoStroke, applyPanZoom } from '$lib/utils/applyPanZoom';
 	import { loadRegionsForApp } from '$lib/utils/loadRegions';
-	import { goto } from '$app/navigation';
-	import { browser } from '$app/environment';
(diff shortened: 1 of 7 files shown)


apps/yapms/src/routes/view/+page.svelte
```diff
@@ -1,44 +1,70 @@
...
-		if (svg !== null) {
-			applyAutoStroke(svg);
-			applyFastPanZoom(svg);
-			setRegionStrokeColor(svg);
-		}
-		loadRegionsForView(node);
-		if ($LoadedMapStore !== null) {
-			loadFromJson($LoadedMapStore);
-		}
+		$effect(() => {
```
Flizzyclone: I don't think effect needs to be here

apps/yapms/src/routes/app/[country]/+layout.svelte
```diff
@@ -1,42 +1,58 @@
...
-			setRegionStrokeColor(svg);
-			loadSidebarTitle(svg);
-			loadSidebarSources(svg);
-			loadMapIdentifier(svg);
-			loadActionGroups(svg);
-		}
-		loadRegionsForApp(node);
-		loadMapFromURL($page.url);
-	}
+		$effect(() => {
```
Comment (Flizzyclone) [inline apps/yapms/src/routes/app/[country]/+layout.svelte]: Same deal

Flizzyclone (approved):
I can't test the ?m bit (I think because I don't have a browserless instance running) but otherwise this LGTM. Loading with `?s=`, `?um=`, and `?m=` causes no issues. 404 still works as before. User maps & saves work.

PR #1477, opened 2025-06-25
Better Regions Table Sort

Flizzyclone:
This PR changes the sort in the regions table to use toLocaleCompare, which fixes issues with districts being out of numerical order.

For example, using the current sort:
image
(1-9 are buried within the other districts, so 2 is only after 19, etc.)

With the new sort:
image

I think this also fixes issues that the sort had with Canadian ridings where ridings that contained characters with accents were sometimes misplaced.


Fish1 (approved):
LGTM

PR #1492, opened 2025-06-28
Modal Toggle Keybinds

Flizzyclone:
This PR adds hotkeys for opening modals. It also adds the ability to select a mode with your keyboard, so you can do a key combination to switch modes (ex: Shift+M+F will switch you to fill)

A lot of switch statement here but I think that's the only way this works.


Fish1 (approved):
LGTM

PR #1514, opened 2025-07-08
Update generate shades color on preset selection

Flizzyclone:
This PR updates the value of the base color used to generate shades when you select a preset slate of colors. Requested by Raya.


Fish1 (approved):
LGTM

PR #1517, opened 2025-07-09
Select for GeoJSON Properties

Flizzyclone:
This PR changes the GeoJsonProperties inputs from text inputs where you would input a property name to a select where you can select from each property name. I use a set to prevent duplicate property names if you are importing multiple shapefiles. My hope is this makes the feature more clear to use (and this will simplify my workflow just a tiny bit!)

image


Fish1 (changes requested):
Just made a couple changes for validation.

Fish1 (approved):
LGTM

PR #1531, opened 2025-07-16
German bug fix + counties

Flizzyclone:
This PR fixes a bug where two candidates on the Bundestag 2025 maps had the same ID, leading to weird behavior. It also adds the German counties map that I made a couple of weeks ago.


Fish1 (approved):
LGTM

PR #1539, opened 2025-07-19
Add preset color selector to the edit candidate modal.

Fish1:
The edit candidate modal didn't have a preset color selector.

I've made a new base modal component so it can be rendered without using a global store.

This should allow the preset color selector modal to be rendered in multiple places with its own state. So closing the preset color modal will go back to the correct previous modal.

### DIFF
--- a/apps/yapms/src/lib/components/modals/NewModalBase.svelte
+++ b/apps/yapms/src/lib/components/modals/NewModalBase.svelte
@@ -36,10 +36,7 @@
 
 	$effect(() => {
 		if (open === true) dialogElement?.showModal();
-	});
-
-	$effect(() => {
-		if (open === false) dialogElement?.close();
+		else if (open === false) dialogElement?.close();
 	});
 </script>
 
--- a/apps/yapms/src/lib/components/modals/NewModalBase.svelte
+++ b/apps/yapms/src/lib/components/modals/NewModalBase.svelte
@@ -11,7 +11,7 @@
 		content = undefined,
 		action = undefined
 	}: {
-		title: string | undefined;
+		title: string;
 		open: boolean;
 		fullWidth?: boolean;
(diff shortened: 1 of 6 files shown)


apps/yapms/src/lib/components/modals/NewModalBase.svelte
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	let contentElement: HTMLDivElement | undefined;
+	let dialogElement: HTMLDialogElement | undefined;
+	let offsetHeight: number | undefined = $state(undefined);
+
+	let isOverflow = $derived(offsetHeight && offsetHeight < (contentElement?.scrollHeight ?? 0));
+
+	$effect(() => {
```
Flizzyclone, line 37: these two $effect functions can be combined into one

apps/yapms/src/lib/components/modals/NewModalBase.svelte
```diff
@@ -0,0 +1,77 @@
...
+		if (open === false) dialogElement?.close();
+	});
+</script>
+
+<dialog class="modal modal-bottom lg:modal-middle" bind:this={dialogElement} onclose={close}>
+	<div class="modal-box flex flex-col w-full" class:!max-w-full={fullWidth}>
+		<h3 class="font-bold">
+			{title ?? ''}
```
Flizzyclone: is there a reason title has to accept undefined? this is a nitpick but it strikes me as odd

Flizzyclone (changes requested):
LGTM for the most part, just two small things on the new base.

apps/yapms/src/lib/components/modals/NewModalBase.svelte
Fish1: Was probably just messing with something at one point. But no reason.

PR #1540, opened 2025-07-20
UK Rework + Expansion

Flizzyclone:
This PR includes a remake of British House of Commons maps for 2019, 2024, and their results to add insets and improve visuals. It also adds maps for Scotland, Northern Ireland, Wales, the London Assembly, European Electoral Regions, and Ceremonial Counties. Historical House of Commons results are also added back to 1997.


Fish1 (approved):
LGTM.

Maybe I missed something, but there is so much to review.

PR #1545, opened 2025-07-27
CRS Conversion

Flizzyclone:
This PR adds a feature when importing a shapefile that allows you to import your .prj file as well to fix invalid coordinate projections that result in maps looking like this: 
image

With the .prj uploaded:
image

reprojectCoordinates is a little weird type wise with the cast to MultiPolygon | Polygon, MultiPolygon's have a 4d array of coordinates and Polygon's have a 3d array which means I have to branch to actually find and project coordinates. Function runs super fast though, on the order of 100s of milliseconds.


Fish1 (changes requested):
The casting "as GeoJSON.Polygon | GeoJSON.MultiPolygon" should be avoided.

Checking the type field of the geometry will deduce that the object has a .coordinates property and be safer.

Image

Fish1:
@Flizzyclone just added some type checking logic for this. Could you validate that this is doing what you expect?

Flizzyclone:
@Fish1 Functionality is working as expected. All looks good besides the failing lint. Thanks for taking care of this.

PR #1549, opened 2025-08-16
Add button to clear locked states or not.

Fish1:
image


Flizzyclone:
What do you think about making this a checkbox that uses a state variable instead? That way you can just toggle it once and hit the clear button like normal.

Fish1:
> What do you think about making this a checkbox that uses a state variable instead? That way you can just toggle it once and hit the clear button like normal.

That sounds better, I'll switch it to that!

Fish1:
> What do you think about making this a checkbox that uses a state variable instead? That way you can just toggle it once and hit the clear button like normal.

Swapped to using a checkbox to clear the locks. Defaulted to checked yes, because that was the original default.

Flizzyclone (approved):

PR #1551, opened 2025-08-21
USA House Redistricting Updates

Flizzyclone:
This PR makes a few changes to the US house maps to reflect redistricting changes. The main house map now uses updated Texas districts, and I have added a secondary map branching from that with redistricted California for if their redistricting ballot amendment passes this November. This secondary map is placed right under in the map card.

image


Flizzyclone:
@Fish1 please review but don't merge yet, I don't want to get this in until maps are passed by the TX state senate, probably tomorrow or Monday.

Fish1 (approved):

PR #1676, opened 2025-09-29
New maps for Missouri redistricting

Flizzyclone:
This PR adds two variants of the US house map, both for if the new Missouri map is upheld. Missouri is a weird situation where there is a relatively straightforward path for implementation of this map to be blocked, and so I do not want to add it to the main map yet. If more "branches" in the road come up like this, I will have to figure something out. Ohio shouldn't branch, and the CA branch will only exist till mid-November.


Fish1:
@Flizzyclone could you rebase your main onto the current main? It's far behind.

Flizzyclone:
@Fish1 done!

Fish1 (changes requested):
The variant of the mo_hb1 map is incorrect. It says blank.

Flizzyclone:
@Fish1 variant fixed

Fish1 (approved):

PR #1743, opened 2025-10-18
Austrian States

Fish1:
Adds Austrian states map.


Flizzyclone (approved):
Added candidate margins and did compression. LGTM

PR #1752, opened 2025-10-22
YAPms Redistricting Council Fair Map

Flizzyclone:
This PR adds the Fair map created by the YAPms Discord Redistricting Council. Adds both the blank map and the map filled in with 2024 presidential results in each district. New map card is under the fantasy section.

### DIFF
--- a/apps/yapms/src/lib/components/mapcard/mapcards/YRCMapCard.svelte
+++ b/apps/yapms/src/lib/components/mapcard/mapcards/YRCMapCard.svelte
@@ -1,0 +1,22 @@
+<script lang="ts">
+	import MapCard from '../MapCard.svelte';
+
+	const links = [
+		{
+			label: 'Fair Map',
+			route: '/app/yrc/fair/2025294/blank'
+		},
+		{
+			label: 'Fair Map (2024 Pres)',
+			route: '/app/yrc/fair/2025294/2024_pres'
+		}
+	];
+</script>
+
+<MapCard
+	name="YAPms Redistricting Council"
+	bg="usa"
+	attribution="Photo by Andy Feliciotti"
+	attributionLink="https://unsplash.com/photos/brown-concrete-building-near-green-trees-during-daytime-6kA9FjqUxhM"
+	{links}
+/>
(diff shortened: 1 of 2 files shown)


Fish1 (changes requested):
Could you minify these maps? They don't look like they need to be 4+ MB.

Flizzyclone:
> Could you minify these maps? They don't look like they need to be 4+ MB.

This is with them run through compression. Size is an artifact of Axel not simplifying the shapefiles themselves. I could theoretically reassemble them from the base shapefiles.

Fish1:
> > Could you minify these maps? They don't look like they need to be 4+ MB.
> 
> This is with them run through compression. Size is an artifact of Axel not simplifying the shapefiles themselves. I could theoretically reassemble them from the base shapefiles.

I can try when I get home tomorrow. But in inkscape if you select a shape, and put it in the edit nodes mode, you can press "ctrl + shift + L" or "ctrl + L" (i forget) and it will try to deleted the unneeded nodes in the shape. But you need to do it one shape at a time.

Flizzyclone:
> > > Could you minify these maps? They don't look like they need to be 4+ MB.
> > 
> > 
> > This is with them run through compression. Size is an artifact of Axel not simplifying the shapefiles themselves. I could theoretically reassemble them from the base shapefiles.
> 
> I can try when I get home tomorrow. But in inkscape if you select a shape, and put it in the edit nodes mode, you can press "ctrl + shift + L" or "ctrl + L" (i forget) and it will try to deleted the unneeded nodes in the shape. But you need to do it one shape at a time.

I'd be curious to see what results you get out of that. I remember it not being good at preserving shapes. I will also see if there is a way to run the algo I use in mapshaper on SVG paths. Not in a rush to get this in either way.

Fish1:
> > > > Could you minify these maps? They don't look like they need to be 4+ MB.
> > > 
> > > 
> > > This is with them run through compression. Size is an artifact of Axel not simplifying the shapefiles themselves. I could theoretically reassemble them from the base shapefiles.
> > 
> > 
> > I can try when I get home tomorrow. But in inkscape if you select a shape, and put it in the edit nodes mode, you can press "ctrl + shift + L" or "ctrl + L" (i forget) and it will try to deleted the unneeded nodes in the shape. But you need to do it one shape at a time.
> 
> I'd be curious to see what results you get out of that. I remember it not being good at preserving shapes. I will also see if there is a way to run the algo I use in mapshaper on SVG paths. Not in a rush to get this in either way.

The ctrl+L in inkscape is really bad, it will make the shapes look weird. But a lot of time it can be covered up with a slightly thicker stroke width on the SVG. But I think it's better than having it be laggy on someones phone or laptop.

Flizzyclone:
Remade with minified shapefiles and about 1.1kb now

PR #1758, opened 2025-10-23
Yukon Territory 2025

Flizzyclone:
The Yukon Territory has an election in ~2 weeks under new maps. Suggestion on the Discord got a half-dozen updates. This PR adds said map. Also reorganizes the Canadian map card a little to pull up maps relevant for 2025/26.


Fish1 (changes requested):
Could you also minify this? The original map is 55 KB, this one is 299 KB. It could probably get to close the same size.

Flizzyclone:
> Could you also minify this? The original map is 55 KB, this one is 299 KB. It could probably get to close the same size.

Yep, down to 57kb.

Fish1 (approved):

PR #1786, opened 2025-11-01
NYC City Council 2025 Map

Flizzyclone:
Since the mayoral election is of intrigue this year, I made this this map a few months ago and never pushed it (also an IRL friend asked). Probably should've done it earlier, oh well. Don't plan on making more city maps for official site.


Fish1 (approved):
LGTM

PR #1797, opened 2025-11-05
Remove Californian Variants of 2026 House

Flizzyclone:
California has resoundingly passed it's redistricting amendment, and therefore we no longer need two separate variants of the house map with it redistricted/not redistricted. This PR sets the previous CA redistricted map to be the default and the MO redistricted map to be the previous CA+MO redistricted. Also removes links to non-CA-redistricted variants from the map card.

Maps with the old CA can still be found by searching for "October 31st, 2025"


Fish1 (approved):
LGTM

PR #1834, opened 2025-11-19
Fork TX Redistricting and reinstitute TX 2024 map

Flizzyclone:
Texas' 2026 map today was struck down by a circuit court in a voting rights lawsuit. This obviously adds a lot of uncertainty to the Texas map and so this PR reverts the main house map to using the 2024 Texas map (as is the current expectation under the law) and creates a separate set of maps including the redistricted map.

I'm not totally convinced merging this PR is the right move - there's a high likelihood the supreme court issues a stay on this via shadow docket within the next couple of days. I err towards the side of merging in order to give people the option of which map to use based on what they think is likely to happen. And it is easy to revert in another PR if a stay is issued.


Fish1 (approved):
LGTM

PR #1838, opened 2025-11-19
New MS and AL State Senate Maps

Flizzyclone:
Alabama today had a judge impose a new State Senate map, making modifications to two districts. Mississippi had a court impose a new State Senate map in May. This PR adds both of them.


Fish1 (approved):
LGTM

PR #1839, opened 2025-11-19
Fix map search bar sizing

Flizzyclone:
A DaisyUI change recently caused the map search bar to shrink. This PR restores it back to full width.


Fish1 (approved):
LGTM

PR #1850, opened 2025-11-22
Redistricted Texas back into base house map, break out "TX Not Redistricted"

Flizzyclone:
SCOTUS basically instantly put a stay on the order that reverted the Texas back to their 2021 map. Assumption has to be at this point that the court is either going to strike down the order or it won't be restored until after the filing deadline.

So, this PR changes the Texas map back to the 2025 map and then breaks out maps for if the 2021 map is put back into use. This is just a rearranging of the maps that were in before this PR to fit the new situation.


Fish1 (approved):
LGTM

PR #1894, opened 2025-12-11
Fix mo_hb1 variant property

Flizzyclone:
This PR fixes a badly set variant property for the MO redistricted house map. It also creates a symlink for the bad variant string to redirect to the correct map, fixing ols saves creating with the error.


Fish1 (approved):
LGTM

PR #1947, opened 2026-01-15
North Dakota and South Carolina '24 state leg maps

Flizzyclone:
Back in 2024, North Dakota and South Carolina had very minor changes to their state legislative maps. I was too lazy to update them then so I am doing it now.


Fish1 (approved):
LGTM

PR #1950, opened 2026-01-17
Short and long name editing

Flizzyclone:
This PR adds inputs in the edit region modal that allow people to change the `short-name` and `long-name` properties for their custom maps without having to go into Inkscape or a text editor. It also enables changes in those properties to be reflected in an exported SVG.


Fish1 (approved):
LGTM

PR #1999, opened 2026-02-07
Remove un-needed assignments

Fish1:
These variables don't need to be assigned because they are assigned later in the control flow. Fixes eslint issue.


Flizzyclone (approved):
LGTM

PR #2045, opened 2026-02-20
Updated VA redistricted maps

Flizzyclone:
The VA Senate made changes to two districts in the Virginia Beach area while considering the redistricting amendment. This PR adds updated maps to reflect those changes now that they have been confirmed by the VA house.


Fish1 (approved):
lgtm

PR #2055, opened 2026-02-23
Automatically center & rotate imported maps when using built-in projection options

Flizzyclone:
This PR adds a option to the custom map dialog that centers and rotates imported maps when using the built-in projection options. 

image

I set the option to be enabled by default as I would define having the map upright and centered as the expected behavior.

This option is not used and not shown when a custom projection is being used - those will have their own rotations and centerings and having the option enabled for custom projections results in wrongly rotated maps.

Example without setting:
image

Example with:
image

### DIFF
--- a/apps/yapms/src/lib/utils/importMap.ts
+++ b/apps/yapms/src/lib/utils/importMap.ts
@@ -1,7 +1,7 @@
 import { ImportedSVGStore } from '$lib/stores/ImportedSVG';
 import { ImportModalStore } from '$lib/stores/Modals';
 import rewind from '@turf/rewind';
-import { geoBounds, geoPath, geoProjection, type GeoRawProjection } from 'd3';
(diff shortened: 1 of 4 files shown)


apps/yapms/src/lib/utils/importMap.ts
```diff
@@ -148,7 +148,18 @@ function geoJsonToSVG(districtShapes: GeoJSON.FeatureCollection) {
 		);
 	}
 
-	const projection = importOptions.projectionFunction().fitSize([width, height], districtShapes);
+	let projection = importOptions.projectionFunction();
+
+	if (importOptions.rotateAndCenter && importOptions.projectionFunction !== proj4ToProjection && projection.rotate !== undefined && projection.center !== undefined) {
```
Flizzyclone: Some projections like Albers USA do not have a defined rotation function

Fish1 (approved):
LGTM

PR #2084, opened 2026-03-05
Mass disable, enable, lock, & unlock

Flizzyclone:
This PR adds new functionality to the Tools modal allowing you to disable, enable, lock, and unlock all regions at once.

massDisableYAPms

To achieve this, I modify ToolButtons to accept a callback and then pass in functions defined in regionActions.

### DIFF
--- a/apps/yapms/src/lib/components/modals/toolsmodal/ToolButton.svelte
+++ b/apps/yapms/src/lib/components/modals/toolsmodal/ToolButton.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { ToolsModalStore } from '$lib/stores/Modals';
 
-	const { text, callback } = $props();
+	const { text, callback }: { text: string; callback: () => void } = $props();
 </script>
 
 <button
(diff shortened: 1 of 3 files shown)


Fish1 (changes requested):
Looks good, but can you just change the ToolButton to use the svelte 5 syntax?

The props should use $props() and the on:click should be onclick.

Flizzyclone:
Yeh sure. Will do tomorrow.

apps/yapms/src/lib/components/modals/toolsmodal/ToolButton.svelte
```diff
@@ -1,16 +1,15 @@
...
-	import type { Writable } from 'svelte/store';
 
-	export let text: string;
-	export let store: Writable<{ open: boolean }>;
-
-	function openToolModal() {
-		$store.open = true;
-		$ToolsModalStore.open = false;
-	}
+	const { text, callback } = $props();
```
Fish1: This just needs type information. Not sure why the linter isn't catching it.

```
const { text, callback }: { text: string, callback: () => void } = $props();
```

Fish1 (changes requested):

apps/yapms/src/lib/components/modals/toolsmodal/ToolButton.svelte
Flizzyclone: Done!

Fish1 (approved):

PR #2085, opened 2026-03-05
Home Rewrite

### LINKED ISSUE #1227: Home Page - Redeisgn
The home page has the terrible grid dots optical illusion. This is most noticeable on light modes. Need a fix to eliminate it.

https://en.wikipedia.org/wiki/Grid_illusion

Image

Flizzyclone:
My apologies for the absolute size of this lad. Most of the change is rewriting map cards into a new format. I'm going to link important files and details about them below. Files with code changes and not data changes are linked below:
- [MapCard.svelte](https://github.com/yapms/yapms/pull/2085/changes#diff-5dde2dfd0442c4fbab10989d34b72cc47bb303ba01d9f20c521600714b552638)
- [MapCardGrid.svelte](https://github.com/yapms/yapms/pull/2085/changes#diff-1fa78457abcd79506830278cbe1bf38068a099185fa40e889a8784262819a958)
- [MapCardLinkGroup.svelte](https://github.com/yapms/yapms/pull/2085/changes#diff-8f880eb1454de71509a923569c500d8c4ab68e3fafc944c41a0d90f2c3485528)
- [MoreMapsModal.svelte](https://github.com/yapms/yapms/pull/2085/changes#diff-8969bfc2e87d4205aff50240aa2f180fe790f09490932d5b3efc09e0a2173bce)
- [styles/global.css](https://github.com/yapms/yapms/pull/2085/changes#diff-67a7cb9fc0a119c3685621c3bce503208e509f56e8e98e3e89e2ff5e85aaa235)
- [types/HomeData.ts](https://github.com/yapms/yapms/pull/2085/changes#diff-d748948c88c1cf4223d8c1574ebc39ee815d884c37cb14900cf85ff442c5eacf)
- [src/routes/+page.svelte](https://github.com/yapms/yapms/pull/2085/changes#diff-f802b2765931d75fadd57fcbf99f28da048f4c95f190f8fff771b6ecd99a4220)

As a general foreword: When I had to make a choice between adding more capability in this PR at the cost of complexity & sometimes obscurity, I chose more capability every time. Therefore, I think the decision of whether to merge this idea at all, even if every issue is ironed out, is not straightforward. To merge is to essentially choose this extra complexity for visual flair + more routes on the home screen at once. I don't think there is a *right* decision, but I wanted to get this idea finished.

## Map card data structure changes
The best place to start with this is looking at the types in [HomeData.ts](https://github.com/yapms/yapms/pull/2085/changes#diff-d748948c88c1cf4223d8c1574ebc39ee815d884c37cb14900cf85ff442c5eacf):

### Groups 

The fundamental change that this PR makes is that map cards are organized into groups of routes instead of an array of routes. A group has four properties:
- `label`: Text used to describe the group, displays above the group's links on the map card, does not show in MoreMapsModal. Special behavior if undefined, which is discussed under "Group label property & showing as many as possible"
- `routes`: An array of routes that fall within the group.
- `showOnCard`: Whether this group's routes will show up on the map card
- `showInModal`: Whether this group's routes will show up in the MoreMapsModal

The base map card component accepts an array of groups as a property. The practice of using the first 3 routes for the map card but not the MoreMapsModal is no longer necessary. The MoreMapsModal will show all routes in the order they are defined within the groups array (i.e the groups array is flattened).

### modalLabel

Routes themselves have also gotten a third property, `modalLabel`. 

When undefined, the map card itself and the MoreMapsModal will use the value of `label` when displaying the route.
When defined, `modalLabel`'s value will show in the MoreMapsModal and `label`'s value will show in the map card.

image

### Attribution
Since map cards now use from 2 to 4 images blended together (or a single 1200x600/2160x600 image, see Yapmeria card), the way attribution is handled is changed. Instead of having both the `attribution` and `attributionLink` properties, a map card now takes a single property `attribution`, an array of type HomeLinkData. These links are then displayed in the MoreMapsModal when you click the attribution icon ([relevant function](https://github.com/yapms/yapms/pull/2085/changes#diff-5dde2dfd0442c4fbab10989d34b72cc47bb303ba01d9f20c521600714b552638R48)):

image

Most additional images used in this PR are in the public domain. The ones that are not are CC-BY 3.0 or CC BY-SA 3.0. We use images under these licenses on the site already.

### Example

A commented example of a map card definition using these new data structures is available in the [ExampleMapCard](https://github.com/yapms/yapms/pull/2085/changes#diff-0d9d3f007ec77bbc8bd6aa77dff082ce7f4e6db27183f569a998f25643aebcd0).

## Map card visual changes

All of the changes made to the data structures for map cards are in service of changing how they look, with emphasis on displaying as many routes as possible.

A map card looks as follows:

image

Card title is shown in the top left, with groups shown across the bottom until the "All Maps" button which opens the MoreMapsModal with all routes for this country. The attribution button is visually unchanged.

### Showing groups

Within a map card, groups are shown with their label in white, centered text at the top and routes below in blue underlined text. When `label` is undefined, no text will show.

Groups will show from left to right in the order that they are defined. **A group must have space to display all of it's routes to be shown.** The number of columns per group (= to ceil[routes/2]) is unlimited.

### Group label property & showing "as many as possible"

I mentioned earlier in this PR that the `label` property for a group has special behavior when undefined. The special behavior is that a group with `label` undefined can partially show, i.e there does not have to be space for the entire group to show up in order for some routes in the group to be shown. Instead, a column of routes will pop in as space becomes available:

linksPopInYAPms

### square & full properties

The base map card component takes in `square` and `full` properties, both booleans. Only one of these properties may be true at the same time, but they can both be false and are by default.

The `square` property makes a card take up two rows in the grid:

image

The `full` property makes a card take up two columns in the grid:

image

### Mobile display

On mobile, up to 4 links will be shown in addition to the "All Maps" button:

image

The 4 links are selected via the [getMobileLinks](https://github.com/yapms/yapms/pull/2085/changes#diff-5dde2dfd0442c4fbab10989d34b72cc47bb303ba01d9f20c521600714b552638R21) map card function, which uses the following rules:
- If 4 groups or greater, show the first link from the first four groups.
- If two groups, show the first two links from the first two groups. Show less if two per group are not available.
- If one group, show the first 4 links from that group, showing less if less than 4 are available.
- If no groups, you messed up. This will return an empty array and show nothing.

Less than 4 links is gracefully handled:

image

## MoreMapsModal changes

The [MoreMapsModal](https://github.com/yapms/yapms/pull/2085/changes#diff-8969bfc2e87d4205aff50240aa2f180fe790f09490932d5b3efc09e0a2173bce) had two changes:
- `$MoreMapsModalStore.title` is now set to include either "Maps" or "Images" as a postfix to accommodate the modal being used for attribution (see [openMoreMapsModal](https://github.com/yapms/yapms/pull/2085/changes#diff-5dde2dfd0442c4fbab10989d34b72cc47bb303ba01d9f20c521600714b552638R43) and [openAttributionModal](https://github.com/yapms/yapms/pull/2085/changes#diff-5dde2dfd0442c4fbab10989d34b72cc47bb303ba01d9f20c521600714b552638R51)).
- Button text will use `modalLabel` if defined.

## Misc
- [global.css](https://github.com/yapms/yapms/pull/2085/changes#diff-67a7cb9fc0a119c3685621c3bce503208e509f56e8e98e3e89e2ff5e85aaa235) was modified to add 3xl and 4xl breakpoints for styling on large displays
- The home page's [+page.svelte](https://github.com/yapms/yapms/pull/2085/changes#diff-f802b2765931d75fadd57fcbf99f28da048f4c95f190f8fff771b6ecd99a4220) has the USA Primaries map card removed, Canada put below Australia, and USA State Senate put under USA State House


Fish1:
Text wrapping is needed for the cards.

image

Fish1:
Also maybe the YAPMS at the top left should be increased in font. I personally think it feels weird for the sub-headers of the countries to be larger than the main logo. I'd say it should be the same size at least.

Maybe just all the font's on the page should be increased in size.

image

Fish1:
I think the areas that contain the links, should also have some padding. They can get too close to the buttons.

image

Fish1:
Also, I think the source code for the image processing script you created should probably be in it's own repository. It's a bit outside the scope of the yapms project for me to critique here. As long as you generate the blended images, that's good.

apps/yapms/src/lib/components/mapcard/MapCard.svelte
```diff
@@ -1,71 +1,114 @@
...
-		attribution: string;
-		attributionLink: string | undefined;
-		links: { label: string; route: string }[];
-	} = $props();
+	export let name: string;
+	export let bg: string;
+	export let attribution: HomeLinkData[];
+	export let groups: HomeGroupData[];
+	export let full: boolean = false;
+	export let square: boolean = false;
```
Fish1: Was there a specific reason to revert this to the svelte4 syntax? Otherwise, this should be reverted back to the svelte 5 syntax.

apps/yapms/src/lib/components/mapcard/MapCardGrid.svelte
```diff
@@ -1,18 +1,17 @@
...
-	const { title = undefined, children }: { title?: string | undefined; children: Snippet } =
-		$props();
+	export let title: string | undefined = undefined;
 </script>
 
-<div class="flex flex-col gap-2">
+<div class="flex flex-col gap-4 w-full items-center">
 	{#if title !== undefined}
-		<div class="text-3xl text-center lg:text-left lg:ml-4">
+		<div class="text-3xl text-center">
```
Fish1: This could also be text-2xl to match the logo being text-2xl

apps/yapms/src/lib/components/mapcard/MapCardLinkGroup.svelte
```diff
@@ -0,0 +1,57 @@
...
+	const nCols = Math.ceil(links.length / 2);
+
+	let columns: HomeLinkData[][] = [];
+	if (label === undefined) {
+		// Breaks array into chunks of two, ex:
+		// [ [links[0], links[1] ], [ links[2] ] ]
+		columns = Array.from({ length: Math.ceil(links.length / 2) }, (_, i) =>
+			links.slice(i * 2, (i + 1) * 2)
+		);
+	}
```
Fish1: This should be svelte 5 syntax.

Also the columns state should be a derived.by and not a state with an effect.

apps/yapms/package.json
```diff
@@ -25,6 +25,7 @@
 		"@sveltejs/kit": "2.55.0",
 		"@sveltejs/vite-plugin-svelte": "6.2.4",
 		"@tailwindcss/postcss": "4.2.2",
+		"@techstark/opencv-js": "4.12.0-release.1",
```
Fish1: This package can be removed.

apps/yapms/package.json
```diff
@@ -52,6 +53,7 @@
 		"prettier-plugin-svelte": "3.5.1",
 		"proj4": "2.20.4",
 		"shapefile": "0.6.6",
+		"sharp": "0.34.5",
```
Fish1: This package can be removed.

apps/yapms/src/lib/components/modals/moremapsmodal/MoreMapsModal.svelte
```diff
@@ -7,12 +7,12 @@
...
 </script>
 
-<ModalBase title="{$MoreMapsModalStore.title} Maps" store={MoreMapsModalStore}>
+<ModalBase title={$MoreMapsModalStore.title} store={MoreMapsModalStore}>
 	<div slot="content">
 		<div class="flex flex-col gap-3 items-center">
 			{#each $MoreMapsModalStore.buttons as button}
-				<a class="btn btn-primary w-2/3 leading-3.5" href={button.route} onclick={close}>
-					{button.label}
+				<a class="btn btn-primary w-2/3 leading-3.5" href={button.route} on:click={close}>
```
Fish1: This should be reverted back to svelte 5 syntax.

apps/yapms/src/lib/components/mapcard/MapCardGrid.svelte
```diff
@@ -1,18 +1,17 @@
 <script lang="ts">
-	import type { Snippet } from 'svelte';
-
-	const { title = undefined, children }: { title?: string | undefined; children: Snippet } =
-		$props();
+	export let title: string | undefined = undefined;
 </script>
```
Fish1, line 4: This should be svelte 5 syntax.

Fish1 (changes requested):
I like how the new home page looks, I like the new images.

But I think the main issue with the code is that a bunch of stuff was reverted back to the svelte 4 syntax. And any new components should continue using svelte 5 syntax. $prop $state etc..

apps/yapms/src/lib/components/mapcard/MapCard.svelte
Flizzyclone: I think I wrote this before we transitioned this component to Svelte 5, but yeah will fix

apps/yapms/src/lib/components/mapcard/MapCardGrid.svelte
```diff
@@ -1,18 +1,17 @@
 <script lang="ts">
-	import type { Snippet } from 'svelte';
-
-	const { title = undefined, children }: { title?: string | undefined; children: Snippet } =
-		$props();
```
Fish1, line 5: Snippets should be brought back here, as slots are the older svelte syntax.

Fish1 (changes requested):

apps/yapms/src/lib/components/mapcard/MapCardLinkGroup.svelte
```diff
@@ -0,0 +1,57 @@
...
+
+	// Breaks array into chunks of two if no label defined, ex:
+	// [ [links[0], links[1] ], [ links[2] ] ]
+	const columns: HomeLinkData[][] = $derived(
+		label === undefined
+			? Array.from({ length: Math.ceil(links.length / 2) }, (_, i) =>
+					links.slice(i * 2, (i + 1) * 2)
+				)
+			: []
+	);
```
Fish1, line 16: Also, you don't need to change these. But I don't think these are actually states. They never change, they are constant values calculated from the props.

If the prop inputs change, a re-render will be triggered.

You only need to define a state variable if a component changes its own state.

PR #2103, opened 2026-03-10
GBR Next Map

Flizzyclone:
This PR creates a "next house of commons" map for Britain, like Canada has, to include the many parties being created from their political fragmentation.


Fish1 (approved):
LGTM

PR #2109, opened 2026-03-11
Arch Chart

Flizzyclone:
This PR adds an arch chart, often used for visualizing parliaments on places like Wikipedia. The chart is generated as an svg via some trigonometry. One half of candidates are placed to the right side, the other half to the left, with tossup in the middle, remaining space:

image
image

Like in the battle chart, leans will appear in the middle for 2 candidates. For more than 2, leans will appear left to right:

image
image

### DIFF
--- a/apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
+++ b/apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
@@ -68,16 +68,11 @@
 
 	const width = $derived(rows * 2 * rowHeight);
 
-	let points = $state<{ radius: number; theta: number }[]>();
-
-	$effect(() => {
-		// Determine point positions
-		const rows = getRowsFromSeats(numSeats);
-
+	const points = $derived.by(() => {
 		let pointIdx = 0;
 		const unsortedPoints = [];
 
-		for (let i = startingRow; i <= rows && pointIdx < numSeats; i++) {
+		for (let i = startingRow; i < rows && pointIdx < numSeats; i++) {
 			let numInRow = Math.round(fillFactor * dotsOnRow(i));
 			numInRow = Math.min(numInRow, numSeats - pointIdx);
 			// Put all remaining seats lost in rounding on the outside row
@@ -100,7 +95,7 @@
 		}
 
 		// Sort large radius to small radius and then small angle to large angle from beginning
-		points = unsortedPoints.sort((a, b) => a.radius - b.radius).sort((a, b) => a.theta - b.theta);
+		return unsortedPoints.sort((a, b) => a.radius - b.radius).sort((a, b) => a.theta - b.theta);
 	});
 
 	function getRowsFromSeats(targetSeats: number) {
--- a/apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
+++ b/apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
@@ -47,6 +47,23 @@
 
 	const rows = $derived(getRowsFromSeats(numSeats));
 
+	const maxForRows = $derived(getSeatsFromRow(rows));
+
+	const startingRow = $derived.by(() => {
+		let starter = 1;
+		let spotsRemoved = 0;
+
+		while (maxForRows-(getSeatsFromRow(starter+1) + spotsRemoved) >= numSeats) {
+			starter++;
+			spotsRemoved += getSeatsFromRow(starter+1);
+		}
+		return starter;
+	});
+
+	const actualSpots = $derived(maxForRows - getSeatsFromRow(startingRow));
+
+	const fillFactor = $derived(numSeats/actualSpots);
+
 	const width = $derived(rows * 2 * rowHeight);
 
 	let points = $state<{ radius: number; theta: number }[]>();
@@ -57,9 +74,18 @@
 
 		let pointIdx = 0;
 		const unsortedPoints = [];
-		for (let i = rows; i > 0 && pointIdx < numSeats; i--) {
-			let numInRow = dotsOnRow(i);
+
+		console.log("max for rows", maxForRows)
+		console.log("starting row", startingRow)
+		console.log("fill factor", fillFactor)
+		console.log("actual spots", actualSpots)
+		console.log("num needed", numSeats)
+
+		for (let i = startingRow; i < rows && pointIdx < numSeats; i++) {
+			let numInRow = Math.round(fillFactor*dotsOnRow(i));
 			numInRow = Math.min(numInRow, numSeats - pointIdx);
+
+			if (i === rows-1) numInRow = numSeats - pointIdx;
 
 			const radius = i * rowHeight;
 			for (let j = 0; j < numInRow && pointIdx < numSeats; j++) {
@@ -89,6 +115,16 @@
 			rows++;
 		}
 		return rows;
+	}
+
+	function getSeatsFromRow(targetRows: number) {
+		let rows = 0;
+		let seats = 0;
+		while (rows < targetRows) {
+			seats += dotsOnRow(rows);
+			rows++;
+		}
+		return seats;
 	}
 
 	function dotsOnRow(nRow: number) {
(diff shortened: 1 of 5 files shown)


Fish1:
Was changing the default from "battle" to "pie" intentional?

apps/yapms/src/lib/stores/Chart.ts
```diff
@@ -4,4 +4,4 @@ import { writable } from 'svelte/store';
 
 export const ChartPositionStore = writable<ChartPosition>('bottom');
 
-export const ChartTypeStore = writable<ChartType>('battle');
+export const ChartTypeStore = writable<ChartType>('pie');
```
Fish1: Fix: I'd say we should leave the battle as the default.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
```diff
@@ -0,0 +1,134 @@
...
+	function getWidthFromSeats(targetSeats: number) {
+		let rows = 0;
+		let seats = 0;
+		while (seats < targetSeats) {
+			seats += dotsOnRow(rows);
+			rows++;
+		}
+		return rows * (2 * rowHeight);
+	}
+
```
Fish1: Nit Pick:
This could just return getRowsFromSeats(targetSeats) * 2 * rowHeight.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
```diff
@@ -0,0 +1,134 @@
...
+	]);
+
+	const numSeats = $derived([...$CandidateCounts.values()].reduce((total, cur) => total + cur, 0));
+
+	const width = $derived(getWidthFromSeats(numSeats));
+
+	let points = $state<{ r: number; theta: number; color: string }[]>();
+
+	$effect(() => {
+		// Determine point positions
```
Fish1: Fix: You could use $derived.by to set the points state, instead of a state that gets altered by an $effect.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
```diff
@@ -0,0 +1,134 @@
...
+		let pointIdx = 0;
+		const unsortedPoints = [];
+		for (let i = rows; i > 0 && pointIdx < numSeats; i--) {
+			let numInRow = dotsOnRow(i);
+			numInRow = Math.min(numInRow, numSeats - pointIdx);
+
+			const radius = i * rowHeight;
+			for (let j = 0; j < numInRow && pointIdx < numSeats; j++) {
+				// Determine position on the unit circle. Add 0.5 to j to rest on the upper semicircle
+				let radians = ((j + 0.5) * Math.PI) / numInRow;
```
Fish1: Very Nit pick: 
I personally think ```((j + 0.5) / numInRow) * Math.PI``` would be a bit easier to understand. Just easier for me to visualize (25 / 50) * Math.PI

Also more nit pick: 
I think breaking out the partial rotation into a variable and adding them together would make it more understandable.
Image

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
```diff
@@ -0,0 +1,134 @@
...
+
+				// If point on the circle will end up on the lower semicircle, bring it back around to the upper semicircle
+				if (radians > Math.PI) {
+					radians += Math.PI;
+				}
```
Fish1: Fix: I'm not sure if this is needed. Like you said in discord, is probably redundant.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
```diff
@@ -0,0 +1,134 @@
...
+
+	$effect(() => {
+		// Determine point positions
+		const rows = getRowsFromSeats(numSeats);
+
+		let pointIdx = 0;
+		const unsortedPoints = [];
+		for (let i = rows; i > 0 && pointIdx < numSeats; i--) {
+			let numInRow = dotsOnRow(i);
+			numInRow = Math.min(numInRow, numSeats - pointIdx);
```
Fish1, line 77: Nit Pick:

I think this does cause a few weird situations...

I don't have a solution for it.

Image

Fish1 (changes requested):
I think most things look good with the math and everything. Just a few things I labeled as fix.

I only nit picked the math because I was playing with it to make sure I understood it.

apps/yapms/src/lib/stores/Chart.ts
Flizzyclone: Whoops. Was not intentional! Reverted to the wrong thing after testing.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
Flizzyclone: Yeah this is not something I have a great solution for either. Drawing the rows in the other direction (out, instead of in) doesn't yield good results:

image

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
Flizzyclone: Removed function entirely, width is now a derived bit of state from rows which is also derived.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
Flizzyclone: Done, and the loop will now read directly from that piece of state instead of having to do a forEach to add the color to the points object.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
Fish1: I think the solution might be to make the circles bigger/smaller.

Flizzyclone: So, a little bit more on this. Basically, the way Wikipedia parliament diagrams solve this problem is allowing the circle's radius to vary. I might be able to figure out a way to do this.

Flizzyclone:
Not re-requesting review for now. Have an idea on how to fix that last nit.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
Flizzyclone: Disregard above comment. Figured out a way to fix this. It involves drawing rows from the inside out, adding an extra row to the outside, removing as many inner rows as possible to create for a nice "hole" at the bottom.

To get the seat distribution in rows equal as to avoid the last row having too little, we maintain a "fill" factor that is multiplied with the maximum number of dots on a row to determine the actual number of dots on that row. 

The fill factor is optimized to be as close as it can be to 1 by the process of removing inner rows.

This all should make it draw almost identical to wikipedia charts.

Flizzyclone:
See my last comment below the last nit, I can do more explanation if needed.

I also added a color transition. Due to coordinates being laid out in polar coordinates, transitioning positions when you change the # of dots looks nuts, so I didn't add that.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
```diff
@@ -0,0 +1,153 @@
...
+
+	const fillFactor = $derived(numSeats / actualSpots);
+
+	const width = $derived(rows * 2 * rowHeight);
+
+	let points = $state<{ radius: number; theta: number }[]>();
+
+	$effect(() => {
+		// Determine point positions
+		const rows = getRowsFromSeats(numSeats);
```
Fish1: I think this state and effect can be done with a derived.by.

Fish1 (changes requested):
I think this looks good to me.

I just think the state and effect should be switch to a $derived.by.

apps/yapms/src/lib/components/charts/archchart/ArchChart.svelte
Flizzyclone: Fixed!

Flizzyclone:
Actually hold on a sec

Flizzyclone:
All good, just removed a redundant computation of the # of rows needed.

Fish1 (approved):

PR #2131, opened 2026-03-22
Hungarian Assembly & Counties

Flizzyclone:
Hungary has elections upcoming in ~3 weeks. This PR adds maps for their parliament (National Assembly) and Counties (ADM1 regions). Map card image is in the public domain.


Fish1 (approved):
LGTM

PR #2155, opened 2026-04-17
Preset Saving from Candidate Models

Flizzyclone:
This PR adds the ability to save a candidates colors as a preset using a button in the candidate modals. Suggested by RedoxLP on Discord.

The way I do action confirmation in the modal is with a tooltip since there's no way to tell the preset has saved on the same screen otherwise.


Fish1 (approved):
LGTM.

Maybe the only thing I'd suggest is making the text change back when the mouse leaves the button.

Image

Also maybe before saving, it could check if the color already exists and block the user from saving twice?

Fish1 (changes requested):
I think the EditCandidateModal also just needs to match the changes on the AddCandidateModal

Fish1 (approved):

PR #2180, opened 2026-04-22
Post-VA redistricting changes

Flizzyclone:
This PR merges both Missouri redistricting and Virginia redistricting into the base 2026 house map, and changes variants to include old maps instead of new. Variants remain as the process to implement both maps is wrapped up in continuing litigation. This pulls us into accordance with what Wikipedia has.


Fish1 (approved):
LGTM

PR #2181, opened 2026-04-22
Subdirectory support for maps

Flizzyclone:
This PR changes map loading to use glob imports, allowing us to have subdirectories within country map directories. This will be especially useful for US maps.

I did not include reorganization for US maps here, so to test this just drag a map into a new folder and try loading it.

I looked at performance here too. Build performance is unaffected on my end, as is page loading performance.

### DIFF
--- a/apps/yapms/src/routes/app/[country]/+layout.svelte
+++ b/apps/yapms/src/routes/app/[country]/+layout.svelte
@@ -16,6 +16,8 @@
 		gotoLoadedMap
 	} from '$lib/stores/LoadedMap';
 
+	let { children } = $props();
+
 	const requestedMap = $derived(page.url.pathname.replace('/app/', '').replaceAll('/', '-'));
 
 	const map = $derived.by(() => {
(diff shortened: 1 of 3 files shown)


apps/yapms/src/routes/app/[country]/+layout.svelte
```diff
@@ -58,15 +65,21 @@
...
-		use:setupMap
-		id="map-div"
-		class="overflow-hidden h-full outline-none"
-		class:insets-hidden={$MapInsetsStore.hidden}
-		class:texts-hidden={$RegionTextsStore.hidden}
-	>
-		{@html map.default}
-	</div>
+	{#if map !== undefined}
+		<div
```
Comment (Fish1) [inline apps/yapms/src/routes/app/[country]/+layout.svelte]: I think this works... but I find it strange to await before checking the undefined. It feels like the check for undefined should happen before the await.

Fish1 (approved):
LGTM. I was looking into the await map, because it looked weird to await something that might be undefined, but I guess it technically works, so it should be fine.

apps/yapms/src/routes/app/[country]/+layout.svelte
Comment (Flizzyclone) [reply apps/yapms/src/routes/app/[country]/+layout.svelte]: Ah yes that makes sense to me. Will switch around

PR #2223, opened 2026-05-07
New TN map + FL action groups fix

Flizzyclone:
This PR adds the new TN house map passed today & fixed Florida action groups on the previous iteration of the map. I've checked the following:
- URL attributes align and thus saves work
- Action groups work for TN
- Old maps updated to include date in title
- Variants are consistent


Fish1 (approved):
LGTM. I would say just try to make these into two different pull requests. That would make it easier to determine what code changes are for what purpose.

PR #2225, opened 2026-05-08
Format All SVGs v2

Fish1:
Formats all SVGs with prettier HTML.

Adds about 30 seconds to the format task.

Made a second PR because I didn't want to deal with the rebase.

- **update prettier config**
- **format svgs**


Flizzyclone (approved):
Can't find any issues with this. LGTM

PR #2249, opened 2026-05-18
Make margin updating immutable

Flizzyclone:
This PR changes margin updates in the edit candidate/edit tossup modals to be immutable, fixing a bug where you would update a candidate's margins and the color boxes on the bottom of the candidate box would not update.

yapmsCandBoxNotUpdating


Fish1 (approved):
This seems to fix the issue. I don't really understand why this change fixes the problem.

PR #2284, opened 2026-06-02
Simulation

Flizzyclone:
This PR adds a simulation section with per-candidate weighted RNG. Weights are even unless modified. This supports simulating regions as a block (as in fill mode) and each bit of region value (as in split mode). The "split regions" toggle controls this and is set when loading a map based on the default-mode attribute.

To support this, I add some logic to preventNonNumericalInput and preventNonNumericalPaste to allow for decimal input and also DefaultModeStore to store the value of the default-mode attribute.


Flizzyclone:
Will fix the lint issues in this tonight

PR #2285, opened 2026-06-02
Louisiana & Alabama 2026 map and Elimination of MO Not Redistricted variant

Flizzyclone:
This PR changes the house map to use Alabama and Louisiana's 2026 maps and gets rid of the MO Not Redistricted variant after their supreme court dismissed every legal challenge.


Flizzyclone:
Marked as draft so I can switch Alabama's map

Fish1 (approved):

PR #2363, opened 2026-06-30
Queensland 2028

Flizzyclone:
Queensland in Australia had their final redistribution report published today, which means a new map is in effect. This PR adds it to the site. It also fixes a typo in the 2024 map.


Flizzyclone:
Adding a new candidate to this in a bit

Fish1 (approved):

PR #2399, opened 2026-07-11
Quebec 2026 V2

Flizzyclone:
This PR updates the Quebec 2026 map to reflect the boundaries the assemblee nationale enacted, overriding the electoral commission. The old map has been marked as the commission map.


Fish1:
@Flizzyclone I don't get it. Why v2? Why isn't the date changing like normal?

Flizzyclone:
@Fish1 the first version never actually went into effect - this is the map that did, on the date specified. the old map is based on an incorrect file from the electoral commission

Fish1:
I think something like blank_v2, blank_fix or blank_fixed would be better then.

Flizzyclone:
@Fish1 sounds good to me. Changed to blank_v2.

Fish1 (approved):

PR #2433, opened 2026-07-24
Rename maps & folders

Flizzyclone:
This PR adds the ability to rename maps and folders:

image

image

I reuse the same modal for both folders and maps


apps/yapms/src/lib/stores/Modals.ts
```diff
@@ -129,6 +129,12 @@ export const ShareDisabledModalStore = writable({
 	open: false
 });
 
+export const RenameSavedMapModalStore = writable({
+	open: false,
+	name: '',
+	onRename: async (_newName: string) => {}
+});
+
```
Fish1, line 137: I think if this had an `id: string` and a `collection: string (or enum)` , this could be a bit more data oriented. 

Then the RenameSavedMapModal could contain the logic for updating the collection.

Then calling the onRename function could just call the onUpdated function.

Fish1 (approved):
LGTM. Otherwise just suggested an alternative way of possibly going about the logic.

PR #2493, opened 2026-09-11
Revert "Missouri First" map and fix /view for maps within a subdir of the country dir

Flizzyclone:
This PR reverts the house map to use the Missouri map from 2022 after a weird last week of litigation

This also fixes a problem with `/view` (and thus, screenshots for Discord) for the 2026 house maps caused by the `/view` route not being able to read maps out of subdirectories of the country directories (in this case, `usa/house-2026`).


Fish1:
You don't have to do it for this PR, but if you could make different PRs for different issues that would be better.

Fish1 (approved):
LGTM, just did linting.

PR #2499, opened 2026-09-13
Fix tooltip not changing after shortName change

Flizzyclone:
PR does what it says on the tin. Good old JS reference BS


Fish1 (approved):
LGTM

